PR #5, opened 2020-03-03
Add RACF support

jasonodonnell:
This adds schema support for IBM RCAF for managing z/OS mainframe accounts.

```bash
$ vault write openldap/config \
    binddn="cn=admin,dc=hashicorp,dc=com" \
    bindpass="admin" \
    url="ldap://0.0.0.0" \
    schema="racf" \
    length=8
Success! Data written to: openldap/config

$ vault write openldap/static-role/hashicorp \
   username='hashicorp' \
   dn='racfid=XB00000,ou=users,o=dev' \
   rotation_period="1d"
Success! Data written to: openldap/static-role/hashicorp

$ vault read openldap/static-role/hashicorp
Key                    Value
---                    -----
dn                     racfid=XB00000,ou=users,o=dev
last_vault_rotation    2020-03-03T14:02:40.594091-05:00
rotation_period        86400s
username               hashicorp

$ vault read openldap/static-cred/hashicorp
Key                    Value
---                    -----
dn                     racfid=XB00000,ou=users,o=dev
last_vault_rotation    2020-03-03T14:02:40.594091-05:00
password               n7kUe8COUdvv494lmrRQjCaYFrSuIqdn2ljYjwOmhqaEXU1a0XLHALb9fk7SPVde
rotation_period        86400s
ttl                    86400s
username               hashicorp
```

### DIFF
--- a/client/fieldregistry.go
+++ b/client/fieldregistry.go
@@ -59,8 +59,8 @@ type fieldRegistry struct {
 	ObjectSID          *Field `ldap:"objectSid"`
 	OrganizationalUnit *Field `ldap:"ou"`
 	PasswordLastSet    *Field `ldap:"passwordLastSet"`
-	RcafPassword       *Field `ldap:"rcafPassword"`
-	RcafAttributes     *Field `ldap:"rcafAttributes"`
+	RCAFPassword       *Field `ldap:"rcafPassword"`
+	RCAFAttributes     *Field `ldap:"rcafAttributes"`
 	UnicodePassword    *Field `ldap:"unicodePwd"`
 	UserPassword       *Field `ldap:"userPassword"`
 	UserPrincipalName  *Field `ldap:"userPrincipalName"`
--- a/client/fieldregistry.go
+++ b/client/fieldregistry.go
@@ -59,8 +59,8 @@ type fieldRegistry struct {
 	ObjectSID          *Field `ldap:"objectSid"`
 	OrganizationalUnit *Field `ldap:"ou"`
 	PasswordLastSet    *Field `ldap:"passwordLastSet"`
-	RCAFPassword       *Field `ldap:"rcafPassword"`
-	RCAFAttributes     *Field `ldap:"rcafAttributes"`
+	RACFPassword       *Field `ldap:"racfPassword"`
+	RACFAttributes     *Field `ldap:"racfAttributes"`
 	UnicodePassword    *Field `ldap:"unicodePwd"`
 	UserPassword       *Field `ldap:"userPassword"`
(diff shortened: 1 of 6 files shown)


client/schema.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/strutil"
+)
+
+// SupportedSchema returns a slice of different OpenLDAP schemas supported
+// by the plugin.  This is used to change the FieldRegistry when modifying
+// user passwords.
+func SupportedSchema() []string {
```
kalafut: Why have this be a function?

path_config.go
```diff
@@ -13,6 +14,7 @@ import (
 const (
 	configPath            = "config"
 	defaultPasswordLength = 64
+	defaultSchema         = "openldap"
```
kalafut: Note that this will break existing beta testers. That's fine (we don't promise backwards compat), but something to note for docs/changelog/etc.

jasonodonnell: Will do, thanks!

client/schema.go
jasonodonnell: I made `SupportedSchema` a function because you can't declare const slices. I suppose it could be a global level variable.  Do you have a preference?

client/fieldregistry.go
```diff
@@ -59,6 +59,8 @@ type fieldRegistry struct {
 	ObjectSID          *Field `ldap:"objectSid"`
 	OrganizationalUnit *Field `ldap:"ou"`
 	PasswordLastSet    *Field `ldap:"passwordLastSet"`
+	RcafPassword       *Field `ldap:"rcafPassword"`
```
kalafut: nit: should these be "RCAFPassword/Attributes"?

client/schema.go
kalafut:
```suggestion
func SupportedSchemas() []string {
```

Would be more accurate.

client/fieldregistry.go
jasonodonnell: Done!

client/schema.go
jasonodonnell: Pushed this in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/5/commits/6008ee92d0209ad00eabaae4018940c2be03d2da

kalafut:
LGTM

jeffsanicola:
I may be mistaken, but isn't it RACF as opposed to RCAF?

jasonodonnell:
@jeffsanicola Doh, yes, thank you!

PR #10, opened 2020-05-12
Fix: Forward rotation requests from performance nodes to the primary

catsby:
Requests to `rotate-role` need to be forwarded from Performance Standby nodes to the Active node. We also return an error if the rotation fails, instead of silently ignoring it. This should fix an issue where calling `rotate-role/*`  appeared to succeed but actually failed, and generated messages in the logs similar to this:

> unable to rotate credentials in rotate-role: error="error writing WAL entry: cannot write to readonly storage"

### DIFF
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -141,14 +141,17 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 		item.Priority = resp.RotationTime.Add(role.StaticAccount.RotationPeriod).Unix()
 	}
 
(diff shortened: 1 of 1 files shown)


jasonodonnell:
@catsby @kalafut Does this same bug apply to `rotate-root`?  There's no forwarding there as well.

catsby:
> Does this same bug apply to rotate-root? There's no forwarding there as well.

We discussed this and determined that ultimately "no" we don't need to do this, but we decided to for good measure. Add in fe9df04

PR #14, opened 2020-06-17
Add response check to prevent panic

jasonodonnell:
A panic occurs if `resp` is a nil pointer in the event of a password rotation failure.  Added a small check if `resp` is nil before trying to extract WAL ID from it.

### DIFF
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -127,7 +127,7 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 	// In create/update of static accounts, we only care if the operation
 	// err'd , and this call does not return credentials
 	item, err := b.popFromRotationQueueByKey(name)
-	if item == nil || err != nil {
+	if err != nil {
 		item = &queue.Item{
 			Key: name,
 		}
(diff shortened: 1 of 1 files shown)


path_rotate.go
```diff
@@ -127,7 +127,7 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 	// In create/update of static accounts, we only care if the operation
 	// err'd , and this call does not return credentials
 	item, err := b.popFromRotationQueueByKey(name)
-	if err != nil {
+	if item == nil || err != nil {
```
kalafut, line 130: How are we hitting a case when `item == nil` but `err == nil`? Looking at https://github.com/hashicorp/vault/blob/e1d21dc601ced5bd925510ee2fe8238cee67781e/builtin/logical/database/rotation.go#L536 it seems like if `item` is ever `nil` we'll definitely have an error. i.e. the code path would never be different than what it is today?

jasonodonnell: Good catch! It was actually due to `resp` being nil and `item` was trying to use a value from it.  Fixed in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/14/commits/e782ca24a52aa477aad736d5ce30a206b8c55fab

jasonodonnell:
Will do @catsby !

PR #16, opened 2020-07-29
Fix AD password rotation

tomhjp:
# Overview

As reported by a customer, rotating passwords in AD using the OpenLDAP plugin didn't work, even when the same config for the AD plugin pointing at the same instance of AD did work.

This is due to the password for AD being stored in the `unicodePwd` field instead of `userPassword`, as documented [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/6e803168-f140-4d23-b2d3-c3a8ab5917d2). The `userPassword` field can be supported too, but it depends on some additional config on the AD server, so `unicodePwd` is the canonical field to update.

To support this, I've added a new supported schema, `"ad"`, for the OpenLDAP plugin, which must be set in the config.

# Contributor Checklist

Docs PR: https://github.com/hashicorp/vault/pull/9619
Manual testing output:

```bash
$ vault read openldap/static-cred/bob
Key                    Value
---                    -----
dn                     CN=Bob Johnson,CN=Users,dc=mycompany,dc=local
last_vault_rotation    2020-07-27T15:33:36.876161+01:00
password               opEOFFDJ1XdRBuwEt2ig2QdPcLRsoXnHpdu3rUmdZcPQlK1ym5mG1nblvbPzZkjC
rotation_period        3m
ttl                    2m43s
username               bob.johnson

ldapwhoami -h x.x.x.x -x -D "CN=Bob Johnson,CN=Users,DC=mycompany,DC=local" -w opEOFFDJ1XdRBuwEt2ig2QdPcLRsoXnHpdu3rUmdZcPQlK1ym5mG1nblvbPzZkjC
u:MYCOMPANY\bob.johnson
```

If you'd like to test this yourself, you'll need a real running instance of AD which I can help set up.

Backwards compatible: ✅

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -166,6 +166,49 @@ func TestUpdatePasswordRACF(t *testing.T) {
 	}
 }
 
+func TestUpdatePasswordAD(t *testing.T) {
+	testPass := "hell0$catz*"
+	encodedTestPass, err := formatPassword(testPass)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	config := emptyConfig()
+	config.BindDN = "cats"
+	config.BindPassword = "dogs"
+
+	conn := &ldapifc.FakeLDAPConnection{
+		SearchRequestToExpect: testSearchRequest(),
+		SearchResultToReturn:  testSearchResult(),
+	}
+
+	dn := "CN=Jim H.. Jones,OU=Vault,OU=Engineering,DC=example,DC=com"
+	conn.ModifyRequestToExpect = &ldap.ModifyRequest{
+		DN: dn,
+	}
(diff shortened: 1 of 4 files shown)


client/client_test.go
```diff
@@ -166,6 +166,49 @@ func TestUpdatePasswordRACF(t *testing.T) {
...
+		LDAP:   &ldapifc.FakeLDAPClient{conn},
+	}
+
+	client := &Client{ldapClient}
+
+	filters := map[*Field][]string{
+		FieldRegistry.ObjectClass: {"*"},
+	}
+
+	newValues, err := GetSchemaFieldRegistry("ad", testPass)
```
jasonodonnell, line 202: It would be good to add a check for `newValues` to ensure the registry is what we expect.

tomhjp: Good point, done

PR #18, opened 2020-08-26
Update api/sdk/vendoring

kalafut:


calvn:
Should we issue a new patch release so that `go get` (which resorts to latest tag if nothing is specified) works properly?

kalafut:
@calvn Cherry pick and release a patch on vault/release-1.5.x?

PR #21, opened 2021-01-13
Add ability to create dynamic credentials in OpenLDAP

pcman312:
# Overview
Adds the ability to create dynamic credentials. These credentials will automatically be removed within the OpenLDAP system after a specified TTL. This handles very similarly to the existing Database engine.

# Design of Change
Dynamic roles require a templated [LDIF](https://en.wikipedia.org/wiki/LDAP_Data_Interchange_Format) in order to create the user within the LDAP system. The LDIF is templated so operators can specify runtime-generated fields (such as the username and password) within the LDIF without knowing them ahead of time. This gives operators a huge amount of flexibility in how a given dynamic user is created.

## Endpoints
Note: All of these endpoints are under `/role` instead of `/roles` to match the existing `/static-role` path
- `POST`/`PUT` `/role/:name` - Creates a dynamic role configuration
  - Parameters:
    - `name` - The name of the role. Specified in the path. Required.
    - `creation_ldif` - The templated LDIF string to use to create the user. Required.
    - `deletion_ldif` - The templated LDIF string to use when removing the user. Required.
    - `rollback_ldif` - The templated LDIF string to use when user creation fails partway through. Optional.
    - `username_template` - Template to use to generate a username. Optional.
    - `default_ttl` - How long a given dynamic user will exist before being removed from the LDAP system. This can be extended using Vault's [lease renewal system](https://www.vaultproject.io/docs/concepts/lease) up but not past the `max_ttl`. Default: Uses the engine's default TTL
    - `max_ttl` - The maximum length a given dynamic user can exist (including extensions of the original lease). Default: Uses the engine's default TTL
- `GET` `/role/:name` - Retrieves a dynamic role configuration
- `LIST` `/role` - Lists all dynamic roles. This will list the names of the roles, not the details of each role.
- `DELETE` `/role/:name` - Deletes a dynamic role configuration

## LDIF
An example LDIF string that can be used in the `creation_ldif` parameter:
```
dn: cn={{.Username}},ou=users,dc=learn,dc=example
objectClass: person
objectClass: top
cn: learn
sn: learn
memberOf: cn=dev,ou=groups,dc=learn,dc=example
userPassword: {{.Password}}
```

~~The `creation_ldif` will be rejected if:~~
~~1. It has multiple entries (denoted by two newlines)~~
~~2. Any of the LDIF entries are modify or deletion commands (denoted by `changetype: modify` or `changetype: delete`~~

These restrictions have been removed since we are allowing for multiple LDIF entries and providing the `rollback_ldif` and `deletion_ldif` templates.

- [x] https://github.com/hashicorp/vault/pull/10692

# Contributor Checklist
- [ ] Docs TBD
- [x] Backwards compatible - No changes to the existing config & static roles
<details>
<summary>Acceptance tests</summary>

```
$ VAULT_PLUGIN_DIR="~/dev/vault/plugins/linux" ./acceptance-tests.bats
   Wed Jan 13 16:04:04 MST 2021 - Starting OpenLDAP server...
   Wed Jan 13 16:04:06 MST 2021 - Bootstrapping OpenLDAP data...
   Wed Jan 13 16:04:16 MST 2021 - OpenLDAP server is running!
   Wed Jan 13 16:04:16 MST 2021 - Starting Vault with plugin directory [/Users/mgolowka/dev/vault/plugins/linux]
   Wed Jan 13 16:04:17 MST 2021 - Waiting for vault to become available...
 ✓ Dynamic Secrets - Read/write role 
 ✓ Dynamic Secrets - List roles 
 ✓ Dynamic Secrets - Credential lifecycle without renewal 
    Wed Jan 13 16:04:44 MST 2021 - Generating credentials...                                                                                            4/4
   Wed Jan 13 16:04:50 MST 2021 - Waiting for a couple of seconds...
   Wed Jan 13 16:04:50 MST 2021 - Renewing lease...
   Wed Jan 13 16:04:50 MST 2021 - Sleeping until after original TTL (1s)...
   Wed Jan 13 16:04:56 MST 2021 - Sleeping until lease expires
 ✓ Dynamic Secrets - Credential lifecycle with renewal 
   Wed Jan 13 16:05:06 MST 2021 - Tearing down containers...
   Wed Jan 13 16:05:06 MST 2021 - Killing container vault-tests...
   Wed Jan 13 16:05:06 MST 2021 - Container vault-tests has stopped
   Wed Jan 13 16:05:06 MST 2021 - Killing container openldap-tests...
   Wed Jan 13 16:05:06 MST 2021 - Container openldap-tests has stopped
   Wed Jan 13 16:05:07 MST 2021 - Teardown complete

4 tests, 0 failures
```
</details>


path_dynamic_creds.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	respData := map[string]interface{}{
+		"username": username,
+		"password": password,
+		"dn":       addReq.DN,
+	}
+	internal := map[string]interface{}{
+		"name":     roleName,
+		"username": username,
+		"password": password,
```
calvn, line 84: Do we need to store the username and password as internal data? AFAIK the data in this map is used for renewal and revocation, but the handlers for those shouldn't need `username` and `password`.

path_dynamic_roles.go
```diff
@@ -0,0 +1,295 @@
...
+			"username": {
+				Type:        framework.TypeString,
+				Description: "Username of the generated account",
+			},
+			"password": {
+				Type:        framework.TypeString,
+				Description: "Password to access the generated account",
+			},
+			"DNs": {
+				Type:        framework.TypeStringSlice,
```
calvn, line 121: Is this correct, or should it be `"dn"` and of type `framework.TypeString`?

path_dynamic_creds.go
```diff
@@ -0,0 +1,242 @@
...
+			return nil, fmt.Errorf("failed to retrieve dynamic role: %w", err)
+		}
+		if dRole == nil {
+			return nil, fmt.Errorf("unable to renew: role does not exist")
+		}
+
+		resp := &logical.Response{
+			Secret: req.Secret,
+		}
+		return resp, nil
```
calvn, line 140: Don't we have to extend/update the `resp.Secret.TTL` (and `resp.Secret.MaxTTL`) value based on the info we got from `dRole`?

path_dynamic_creds.go
```diff
@@ -0,0 +1,242 @@
...
+	RoleName    string
+}
+
+func generateUsername(req *logical.Request, role *dynamicRole) (string, error) {
+	if role.UsernameTemplate == "" {
+		randStr, err := base62.Random(20)
+		if err != nil {
+			return "", err
+		}
+		username := fmt.Sprintf("v_%s_%s_%s_%d", req.DisplayName, role.Name, randStr, time.Now().Unix())
```
calvn, line 185: Is this default format documented elsewhere?

path_dynamic_roles.go
```diff
@@ -0,0 +1,295 @@
...
+// convertToDuration all keys in the data map into time.Duration objects. Keys not found in the map will be ignored
+func convertToDuration(data map[string]interface{}, keys ...string) error {
+	merr := new(multierror.Error)
+	for _, key := range keys {
+		val, exists := data[key]
+		if !exists {
+			continue
+		}
+
+		switch v := val.(type) {
```
calvn, line 174: You could use `parseutil.ParseDurationSecond` here instead.

calvn (commented):
Looking great! Left a few comments/suggestions.

client/client.go
```diff
@@ -143,3 +143,47 @@ func shouldTryLastPwd(lastPwd string, lastBindPasswordRotation time.Time) bool {
...
+func (c *Client) Add(cfg *Config, req *ldap.AddRequest) error {
+	if req == nil {
+		return fmt.Errorf("invalid request: request is nil")
+	}
+	if req.DN == "" {
+		return fmt.Errorf("invalid request: DN is empty")
+	}
+	conn, err := c.ldap.DialLDAP(cfg.ConfigEntry)
+	if err != nil {
+		return err
```
kalafut, line 156: Might want to wrap these e.g. `fmt.Errorf("error adding user: %w", err)` for better log context awareness, since the pattern is so similiar to `Del()`

dynamic_role.go
```diff
@@ -0,0 +1,61 @@
...
+func retrieveDynamicRole(ctx context.Context, s logical.Storage, roleName string) (*dynamicRole, error) {
+	entry, err := s.Get(ctx, path.Join(dynamicRolePath, roleName))
+	if err != nil {
+		return nil, err
+	}
+	if entry == nil {
+		return nil, nil
+	}
+
+	var result *dynamicRole
```
kalafut, line 32: I think this should this be `dynamicRole` since you're sending in `&result` below. No allocation is occurring here (though I think DecodeJSON _might_ do some magic allocation, I don't like relying on that).

path_dynamic_creds.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("failed to apply creation_ldif template: %w", err)
+	}
+
+	// Parse the raw LDIF & run it against the LDAP client
+	entries, err := ldif.Parse(createLDIF)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse generated LDIF: %w", err)
+	}
+
+	addReq := getAddRequest(entries.Entries[0])
```
kalafut, line 70: Will `ldif.Parse()` return an error if there are no entries, or do we need to guard against that here?

client.go
```diff
@@ -3,11 +3,15 @@ package openldap
 import (
 	"fmt"
 
+	"github.com/go-ldap/ldap/v3"
+
```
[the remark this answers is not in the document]
pcman312: Done

path_dynamic_creds.go
pcman312: This has been changed to include the `deletion_ldif` as well as all of the dynamic template data so the deletion template can receive all of the same data the creation template does.

path_dynamic_roles.go
pcman312: At the time this was correct, however since we've changed it to allow for multiple statements this has been changed back to `DNs` as a string slice. Each element in the slice corresponds to the LDIF statement run.

path_dynamic_creds.go
pcman312: I don't think we do. I have little experience with lease TTLs, but from my testing this doesn't appear to be needed:

```
$ vault write openldap/role/testrole creation_ldif=... deletion_ldif=... default_ttl=1m max_ttl=5m


$ # Credentials command removed here for brevity since I have a bunch of inline stuff going on with it, but the output is:
Lease ID: openldap/cred/testrole/82qMbGXOHKlCb2o3cRXN6els
DN: cn=v_token_testrole_HoL1VANHfN_1611097161,ou=users,dc=learn,dc=example
Password: JMKxJBm4yPUHjxeE7zGfxV0DnfPBnnjwudNRNdUNBqtWRcwb7H5Pj5xr75Ty0FZX
User found


$ curl -s --header "X-Vault-Token: $(vault token lookup -format=json | jq -r .data.id)" -X PUT --data "{\"lease_id\":\"${LEASE_ID}\"}" $VAULT_ADDR/v1/sys/leases/lookup | jq
{
  <snip>
  "data": {
    "expire_time": "2021-01-19T16:00:21.812546-07:00",
    "id": "openldap/cred/testrole/82qMbGXOHKlCb2o3cRXN6els",
    "issue_time": "2021-01-19T15:59:21.812545-07:00",
    "last_renewal": null,
    "renewable": true,
    "ttl": 53
  },
  <snip>
}


$ $ vault lease renew -increment=5m "${LEASE_ID}"
WARNING! The following warnings were returned from Vault:

  * TTL of "5m" exceeded the effective max_ttl of "4m31s"; TTL value is capped
  accordingly

Key                Value
---                -----
lease_id           openldap/cred/testrole/82qMbGXOHKlCb2o3cRXN6els
lease_duration     4m31s
lease_renewable    true


$ curl -s --header "X-Vault-Token: $(vault token lookup -format=json | jq -r .data.id)" -X PUT --data "{\"lease_id\":\"${LEASE_ID}\"}" $VAULT_ADDR/v1/sys/leases/lookup | jq
{
  <snip>
  "data": {
    "expire_time": "2021-01-19T16:04:21.189962-07:00",
    "id": "openldap/cred/testrole/82qMbGXOHKlCb2o3cRXN6els",
    "issue_time": "2021-01-19T15:59:21.812545-07:00",
    "last_renewal": "2021-01-19T15:59:50.189962-07:00",
    "renewable": true,
    "ttl": 266
  },
  <snip>
}
```

path_dynamic_creds.go
pcman312: I was planning on including example outputs in the documentation, however I don't plan on explicitly including the format as a guaranteed part of the API. This gives us the potential flexibility to change it and matches the behavior of the database engine.

path_dynamic_roles.go
pcman312: Ah good catch. I forgot that we had that package. Updated

client/client.go
pcman312: The context is usually added by the caller rather than inside the function otherwise it's easy to end up with `error adding user: error adding user: unable to dial LDAP server`

dynamic_role.go
pcman312: The reason for a pointer here isn't for memory allocation, it's so the ownership of the role (and thus the ability to change the object) is passed to the caller. That being said, I changed `&result` to `result` in the `DecodeJSON` call since it's already a pointer.

path_dynamic_creds.go
pcman312: `ldif.Parse()` does not return an error on an empty string. However, I put a protection in place in the `assertValidLDIFTemplate` function to ensure there's at least one entry within `creation_ldif` and `deletion_ldif`. Additionally, since I've changed this to allow for more than one entry, this particular line doesn't exist any longer so the index out of bounds error won't happen here.

path_dynamic_creds.go
calvn: I believe that the update on the `TTL` and `MaxTTL` values is to pass the latest values from the role entry to the renewal logic if they got updated. So if you were to update the role's `ttl` (or `max_ttl`) to say 1m right before that renewal call, it would cap it accordingly.

dynamic_role.go
kalafut: `result` is still a nil pointer though, which will cause json unmarshalling to panic. So we need either:

```
var result = new(dynamicRole)
```

or

```
var result dynamicRole
...DecodeJSON(&result)...
```

path_dynamic_creds.go
calvn: I'm not quite getting your comments. AFAIK the internal data map is only really useful for lease-related logic (i.e. renew/revoke), but I don't see those entries being used on `secretCredsRenew` or `secretCredsRevoke` when we're accessing the values in this map.

path_dynamic_roles.go
calvn: Should we change `"dn"` [in here](https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/21/files/d52a826c87f68ba2669287a3044b33da78cbcc60#diff-d5b65bfb287fcee080448ea1cbe9ba3f5af4f06b56b059bc2d0a59103f48a407R79) to `"DNs"` then?

client/client.go
kalafut: 👍

path_dynamic_creds.go
kalafut: Sgtm

path_dynamic_creds.go
pcman312: `username` and `password` are not top level fields any longer. `deletion_ldif` is being set in the internal data and used during `secretCredsRevoke` along with `template_data` which contains all of the same information that is applied to the `creation_ldif` template during credential creation.

path_dynamic_roles.go
pcman312: [It has already been changed](https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/21/files#diff-d5b65bfb287fcee080448ea1cbe9ba3f5af4f06b56b059bc2d0a59103f48a407R76) - Your link is to a subset of changes that don't include this particular change.

path_dynamic_creds.go
pcman312: Oh I see what you're saying - if the role itself is modified while leases are active. I'll take a look into this.

dynamic_role.go
pcman312: You are correct, I've updated accordingly. I'm very confused though since this was working previously and I can't figure out why it was working instead of erroring...

path_dynamic_creds.go
calvn: Argh, you're right on both counts -- I was looking at an outdated diff in my other tab.

path_dynamic_roles.go
calvn: I'm not a fan of the capitalized field `"DNs"` since we tend to return snake cased field names, but I also don't have a better suggestion for the name at this point in time.

path_dynamic_creds.go
pcman312: Done. This looks like it works as it should now. The only thing to note is that if I generate creds from a role, then update the role to have a `max_ttl` less than the current elapsed time the existing credentials aren't immediately revoked. They continue to exist until the original expiration. Is this our standard behavior? This would require operators to revoke existing leases (or wait for the original TTL). Or should there be an extra step somewhere to auto-revoke?

path_dynamic_roles.go
pcman312: I completely agree. Unfortunately we have a name conflict with `DNS` :(

We could expand to the full name: `distinguished_names`

pcman312: Updated to `distinguished_names`

path_dynamic_roles.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode request: %w", err)
+	}
+
+	err = validateDynamicRole(dRole)
+	if err != nil {
+		return nil, err
+	}
+
+	dRole.CreationLDIF = decodeBase64(dRole.CreationLDIF)
```
calvn, line 160: Does this mean that the LDIFs will always be provided as base64 encoded values? Your unit and acceptance tests seem to be able to provide them as non-encoded strings, but I'm a bit puzzled by how this is working if `base64.StdEncoding.DecodeString` is always being called on a non-empty string.

client/client.go
```diff
@@ -143,3 +145,121 @@ func shouldTryLastPwd(lastPwd string, lastBindPasswordRotation time.Time) bool {
...
+			// Skip entries that are nil since they don't indicate an error in execution. Since these entries
+			// are usually coming from an ldif parse, this should generally not happen so it's mainly to
+			// protect against developers from screwing up and creating a panic due to a nil reference
+			continue
+		}
+		var err error
+		switch {
+		case entry.Entry != nil:
+			addReq := coerceToAddRequest(entry.Entry)
+			err = errorf("failed to run AddRequest", conn.Add(addReq))
```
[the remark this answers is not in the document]
pcman312: Done. Good catch. I'm not particularly fond of the `errorf` function, but I also dislike putting the same conditional in each `case` statement.

path_dynamic_roles.go
pcman312: No, they are optionally base64 encoded.

This works because `base64.StdEncoding.DecodeString` will error if the string in question is not base64 encoded. A plain LDIF string cannot be base64 decoded so it will fail. That error will result in `decodeBase64` returning the original string. It's pretty much the equivalent of "if string is base64 encoded, decode it and return, otherwise return the original" except the conditional and the decoding is happening at the same step.

I'll add a comment on the `decodeBase64` function to try to make this clearer in the code.

client/client.go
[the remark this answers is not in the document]
calvn: Yeah that's fair and I'm okay with that as well.

path_dynamic_roles.go
calvn: Oh I didn't realize that we were ignoring the error and returning `str` if so.

calvn: I know that we call `assertValidLDIFTemplate` down below, but I wonder if we should `len` check `dRole.CreationLDIF` early and return and better error here indicating that the creation LDIF is required.

calvn: 🤦 It's handled under `validateDynamicRole`

pcman312: Looking at this again, I think I'm going to shuffle this around a little bit:
1. Move the base64 decoding above the validation
2. Move the LDIF template assertion into `validateDynamicRole` so all of the validation logic is in one function

path_dynamic_roles.go
```diff
@@ -0,0 +1,347 @@
...
+
+func validateDynamicRole(dRole *dynamicRole) error {
+	merr := new(multierror.Error)
+
+	merr = multierror.Append(merr,
+		firstError(
+			errIfTrue(dRole.CreationLDIF == "", "missing creation_ldif"),
+			errWrap(assertValidLDIFTemplate(dRole.CreationLDIF), "invalid creation_ldif: %w"),
+		),
+	)
```
calvn, line 180: This feels like a bit too much error-handling magic without having to dive into what each of these helpers are doing.

Can we simplify the logic here to something like this?
```
var creationErr error
if dRole.CreationLDIF == "" {
  creationErr = errors.New("missing creation_ldif")
}
if creationErr == nil {
    if err := assertValidLDIFTemplate(dRole.CreationLDIF); err != nil {
      creationErr = fmt.Errorf("invalid creation_ldif: %w", err)
    }
}
merr = multierror.Append(merr, creationErr) 
```

pcman312: I disagree that this will be more readable and more maintainable than other options. This code would need to be copy-pasta'd several times with small changes between them. This will result in future maintainability problems because copy-pasta of large blocks with minor changes is prone to human error.

I've changed this to a fail fast model so we can get this PR merged.

path_dynamic_creds.go
calvn: That's an interesting case, and it's one where I'm not sure whether the correct thing should be to immediately revoke or simply let it cap to its current remaining TTL. I think the behavior that exists today is on the latter, which is what you're observing.

path_dynamic_creds.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func getString(m map[string]interface{}, key string) (string, error) {
+	if m == nil {
+		return "", fmt.Errorf("nil map")
+	}
+
+	val, exists := m[key]
+	if !exists {
+		return "", nil
```
calvn, line 246: Should this be returning an error? I don't know if it's okay to pass in an empty string to `b.executeLDIF` in [here](https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/21/files#diff-d5b65bfb287fcee080448ea1cbe9ba3f5af4f06b56b059bc2d0a59103f48a407R183).

path_dynamic_roles.go
calvn: I'm good with this approach since that's what we default for these types of checks and it keeps the logic clean.

calvn (approved):
Left a few minor comments, but looks good otherwise!

path_dynamic_creds.go
pcman312: Good catch. I think this function should not return an error so it can distinguish between "not found" and "invalid type" or other error. I've added an additional check in `secretCredsRevoke` to check if the template is empty. This should only happen if there's a screw up on the developer's part since the `deletion_ldif` should already exist by this point.

path_dynamic_creds.go
```diff
@@ -0,0 +1,275 @@
...
+		"template_data": templateData,
+	}
+	resp := b.Secret(secretCredsType).Response(respData, internal)
+	resp.Secret.TTL = dRole.DefaultTTL
+	resp.Secret.MaxTTL = dRole.MaxTTL
+
+	return resp, nil
+}
+
+func encodeUTF16LE(str string) (string, error) {
```
calvn, line 94: Nit: can we move this to be on L243 or so, so that it's closer to where it's being called?

pcman312: Done

path_dynamic_roles.go
```diff
@@ -0,0 +1,317 @@
...
+				"default_ttl": {
+					Type:        framework.TypeDurationSecond,
+					Description: "Default TTL for dynamic credentials",
+				},
+				"max_ttl": {
+					Type:        framework.TypeDurationSecond,
+					Description: "Max TTL a dynamic credential can be extended to",
+				},
+			},
+			ExistenceCheck: b.pathDynamicRoleExistenceCheck,
```
austingebauer, line 63: Wondering if the existence check is needed here? It looks like a write is processed the same in `pathDynamicRoleCreateUpdate()` whether the role exists or not.

path_dynamic_creds.go
```diff
@@ -0,0 +1,275 @@
...
+		DisplayName:           req.DisplayName,
+		RoleName:              roleName,
+		IssueTime:             now.Format(time.RFC3339),
+		IssueTimeSeconds:      now.Unix(),
+		ExpirationTime:        exp.Format(time.RFC3339),
+		ExpirationTimeSeconds: exp.Unix(),
+	}
+	dns, err := b.executeLDIF(config.LDAP, dRole.CreationLDIF, templateData, false)
+	if err != nil {
+		// Creation failed, attempt a rollback if one is specified
```
austingebauer, line 63: Nit: It wasn't immediately obvious to me that executing the LDIF could result in multiple operations. Perhaps a small comment to make it more obvious to the next reader why the rollback is needed.

path_dynamic_roles.go
pcman312: It is needed, but not for the create/update endpoint. It was needed for one of the other non-endpoints. I don't recall exactly which one off hand (renewal? revocation?). I remember it was giving me issues if I didn't specify it.

austingebauer (approved):
LGTM

path_dynamic_roles.go
```diff
@@ -0,0 +1,317 @@
...
+				},
+				logical.CreateOperation: &framework.PathOperation{
+					Callback: b.pathDynamicRoleCreateUpdate,
+				},
+				logical.ReadOperation: &framework.PathOperation{
+					Callback: b.pathDynamicRoleRead,
+				},
+				logical.DeleteOperation: &framework.PathOperation{
+					Callback:                    b.pathDynamicRoleDelete,
+					ForwardPerformanceStandby:   true,
```
catsby, line 76: Is there a reason `ForwardPerformanceStandby` and `ForwardPerformanceSecondary` are only on the `DeleteOperation`?

path_dynamic_roles.go
```diff
@@ -0,0 +1,317 @@
...
+	err = mapstructure.WeakDecode(rawData, dRole)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode request: %w", err)
+	}
+
+	dRole.CreationLDIF = decodeBase64(dRole.CreationLDIF)
+	dRole.RollbackLDIF = decodeBase64(dRole.RollbackLDIF)
+	dRole.DeletionLDIF = decodeBase64(dRole.DeletionLDIF)
+
+	err = validateDynamicRole(dRole)
```
catsby, line 159: It looks like in createupdate here that we don't load the role from storage, so if we want to change the _just_ `default_ttl` for a role, I need to re-supply all the parameters again?

catsby (commented):
Question on updating roles

path_dynamic_roles.go
pcman312: In the current form yes. I originally wanted to keep the code simpler by using `mapstructure` to decode the request data. I talked with @catsby offline and was able to get this to do partial updates without having specify each field individually here while still being able to differentiate between "value not set" and "delete this value".

path_dynamic_roles.go
pcman312: Not specifically. I started by copy-pasting the static role definitions which only have these set on `DeleteOperation`. My knowledge of these fields is pretty limited - should they be set on all of the write operations?

path_dynamic_creds.go
pcman312: Added a comment block to `executeLDIF` to hopefully make this clearer.

path_dynamic_roles.go
catsby: I don't believe this is needed at all here. The general rule for using `ForwardPerformanceStandby` and `ForwardPerformanceSecondary` is if the method interacts and modifies an external system, e.g. rotating credentials for a root account or something. Docs: 

- https://github.com/hashicorp/vault/blob/22008eefb69cbd6ad88ec620f83b4f8aaa59b3bc/sdk/framework/path.go#L156-L178

pcman312: Gotcha - that makes sense. Thanks for the info! I've updated the PR to remove these.

path_dynamic_roles.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	roleName := data.Get("name").(string)
+	dRole, err := retrieveDynamicRole(ctx, req.Storage, roleName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to look for existing role: %w", err)
+	}
+	if dRole == nil {
+		dRole = &dynamicRole{}
+	}
```
calvn, line 154: Does this mean that you can update a role that doesn't exist? I think that there needs to be a check here such that if this is an UpdateOperation but the role doesn't exist, we return an error. This is to prevent a token with `update` permission but not `create` permission from creating new roles.

pcman312: Ah good catch. Updated.

PR #23, opened 2021-02-16
A bunch of little fixes

pcman312:
- Renames `/cred` -> `/creds`
- Removes `prefix` argument from static creds endpoint (it isn't supported & was creating problems with the `path-help` command on Vault)
- Fixes creation error not showing up if a rollback failure occurs during dynamic credential creation

### DIFF
--- a/path_dynamic_roles.go
+++ b/path_dynamic_roles.go
@@ -18,8 +18,8 @@ import (
 const (
 	secretCredsType = "creds"
 
-	dynamicRolePath = "role/"
-	dynamicCredPath = "cred/"
+	dynamicRolePath = "roles/"
+	dynamicCredPath = "creds/"
 )
 
 func (b *backend) pathDynamicRoles() []*framework.Path {
--- a/path_dynamic_roles.go
+++ b/path_dynamic_roles.go
@@ -18,7 +18,7 @@ import (
 const (
 	secretCredsType = "creds"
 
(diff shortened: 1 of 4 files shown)


path_dynamic_roles.go
```diff
@@ -18,8 +18,8 @@ import (
 const (
 	secretCredsType = "creds"
 
-	dynamicRolePath = "role/"
-	dynamicCredPath = "cred/"
+	dynamicRolePath = "roles/"
```
calvn, line 21: Consider reverting this back to `role/`

calvn (commented):
One request on the `role/` API path, otherwise 👍

path_dynamic_roles.go
pcman312: Done

scripts/acceptance-tests.bats
```diff
@@ -183,11 +183,11 @@ teardown() {
   vault delete openldap/config
 
   # Remove any roles that were created so they don't bleed over to other tests
-  output=$(vault list -format=json openldap/role || true) # "or true" so it doesn't show an error if there are no roles
+  output=$(vault list -format=json openldap/roles || true) # "or true" so it doesn't show an error if there are no roles
```
calvn, line 186: You'd have to revert these paths as well 😄

calvn (approved):

PR #28, opened 2021-09-10
Fix early rotation for roles with WALs, handle multiple WALs per role

tomhjp:
# Early rotation of credentials

## Bug and repro description

When the plugin finds a WAL for a role on initialisation, it (understandably) assumes it needs to rotate the password for the role ASAP. However, the plugin also assumes there can be at most 1 WAL per role at any time, and that's not true. We can have arbitrarily old and stale WALs in storage generated by failures during manual attempts to set up or rotate a role. These only ever get a chance to be cleaned up one at a time for each initialisation the plugin goes through. As a result, we can see early password rotations occur when the plugin runs initialisation.

To illustrate, here is a scenario to reproduce an early rotation:

```bash
vault write openldap/config \
    url=ldaps://vault-ad-test.example.com\
    binddn="CN=vault-ad-admin,CN=Users,DC=example,DC=com" \
    bindpass="some-password" \
    schema=ad

# Ensure the AD server is down so that this command fails
# No storage entry gets created for the role itself, but it will leave a WAL behind in storage for the role anyway
vault write openldap/static-role/mary \
    dn="CN=mary.smith,CN=Users,DC=example,DC=com" \
    username="mary.smith" \
    rotation_period="48h"

# To observe the dangling WAL, check the sys/raw endpoint, which should return a single WAL
MOUNT_UUID=$(vault read -format=json sys/mounts | jq -r '.data["openldap/"].uuid')
vault list sys/raw/logical/$MOUNT_UUID/wal

# Now ensure the AD server is up again
# This command should succeed, in which case create and then quickly delete a WAL, but leave the original WAL untouched
vault write openldap/static-role/mary \
    dn="CN=mary.smith,CN=Users,DC=example,DC=com" \
    username="mary.smith" \
    rotation_period="48h"

# Still 1 WAL
vault list sys/raw/logical/$MOUNT_UUID/wal

# Read the creds
vault read openldap/static-cred/mary

# Reload the plugin, and on initialisation, it will see the first WAL we created and immediately rotate the role, even though it's already been rotated very recently.
vault write sys/plugins/reload/backend mounts=openldap/

# Read the creds again, and you should see they've been rotated
vault read openldap/static-cred/mary
# WAL has been cleared
vault list sys/raw/logical/$MOUNT_UUID/wal
```

## Fix

There are a few meaningful changes:

1. Logging updates
2. Ensure a maximum of 1 WAL per role. There are some areas where we lose track of an existing WAL ID and end up creating an extra WAL for the role instead
3. Use the new password stored in the WAL when available. Previously we generated a new password on every attempt. This ensures we perform fewer rotations when we retry after failed attempts.
4. Delete stale WALs from storage as we load them. Without the above bug fix, this would be a pretty effective mitigation, but as it is, the main motivation now is to avoid a resource leak.
5. Discard WALs that have a last rotation time of 0. They should only be possible when they exist as a hangover from a previous failed attempt to create a role in Vault, and so we should be able to always safely ignore them.

### DIFF
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -24,12 +24,12 @@ func (b *backend) pathRotateCredentials() []*framework.Path {
 			Fields:  map[string]*framework.FieldSchema{},
 			Operations: map[logical.Operation]framework.OperationHandler{
 				logical.UpdateOperation: &framework.PathOperation{
-					Callback:                    b.pathRotateCredentialsUpdate,
+					Callback:                    b.pathRotateRootCredentialsUpdate,
 					ForwardPerformanceStandby:   true,
 					ForwardPerformanceSecondary: true,
 				},
 				logical.CreateOperation: &framework.PathOperation{
-					Callback:                    b.pathRotateCredentialsUpdate,
+					Callback:                    b.pathRotateRootCredentialsUpdate,
 					ForwardPerformanceStandby:   true,
 					ForwardPerformanceSecondary: true,
 				},
@@ -64,7 +64,7 @@ func (b *backend) pathRotateCredentials() []*framework.Path {
 	}
 }
 
-func (b *backend) pathRotateCredentialsUpdate(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+func (b *backend) pathRotateRootCredentialsUpdate(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	if _, hasTimeout := ctx.Deadline(); !hasTimeout {
 		var cancel func()
 		ctx, cancel = context.WithTimeout(ctx, defaultCtxTimeout)
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -149,6 +149,8 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 		}
 	} else {
 		item.Priority = resp.RotationTime.Add(role.StaticAccount.RotationPeriod).Unix()
+		// Clear any stored WAL ID as we must have successfully deleted our WAL to get here.
+		item.Value = ""
 	}
 
 	// Add their rotation to the queue. We use pushErr here to distinguish between
@@ -159,9 +161,16 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 		return nil, pushErr
 	}
 
+	if err != nil {
+		return &logical.Response{
+			Warnings: []string{"unable to finish rotating credentials; retries will " +
+				"continue in the background but it is also safe to retry manually"},
+		}, err
+	}
+
 	// We're not returning creds here because we do not know if its been processed
 	// by the queue.
-	return nil, err
+	return nil, nil
 }
 
 // rollBackPassword uses naive exponential backoff to retry updating to an old password,
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -162,10 +162,8 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 	}
 
 	if err != nil {
-		return &logical.Response{
-			Warnings: []string{"unable to finish rotating credentials; retries will " +
-				"continue in the background but it is also safe to retry manually"},
-		}, err
+		return nil, fmt.Errorf("unable to finish rotating credentials; retries will "+
+			"continue in the background but it is also safe to retry manually: %w", err)
 	}
 
 	// We're not returning creds here because we do not know if its been processed
--- a/path_rotate.go
+++ b/path_rotate.go
@@ -133,10 +133,14 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 		}
 	}
 
-	resp, err := b.setStaticAccountPassword(ctx, req.Storage, &setStaticAccountInput{
+	input := &setStaticAccountInput{
 		RoleName: name,
 		Role:     role,
-	})
+	}
+	if walID, ok := item.Value.(string); ok {
+		input.WALID = walID
+	}
+	resp, err := b.setStaticAccountPassword(ctx, req.Storage, input)
 	if err != nil {
 		b.Logger().Warn("unable to rotate credentials in rotate-role", "error", err)
 		// Update the priority to re-try this rotation and re-add the item to
--- a/path_static_roles.go
+++ b/path_static_roles.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/helper/locksutil"
 	"github.com/hashicorp/vault/sdk/logical"
@@ -146,6 +147,27 @@ func (b *backend) pathStaticRoleDelete(ctx context.Context, req *logical.Request
 		return nil, err
 	}
 
+	walIDs, err := framework.ListWAL(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+	for _, walID := range walIDs {
+		var merr *multierror.Error
+		wal, err := b.findStaticWAL(ctx, req.Storage, walID)
+		if err != nil {
+			merr = multierror.Append(merr, err)
+			continue
+		}
+		if wal != nil && name == wal.RoleName {
+			err = framework.DeleteWAL(ctx, req.Storage, walID)
+			if err != nil {
+				merr = multierror.Append(merr, err)
+			}
+		}
+
+		return nil, merr.ErrorOrNil()
+	}
+
 	return nil, nil
 }
 
@@ -230,6 +252,15 @@ func (b *backend) pathStaticRoleCreateUpdate(ctx context.Context, req *logical.R
 			Role:     role,
 		})
 		if err != nil {
+			if resp != nil && resp.WALID != "" {
+				walDeleteErr := framework.DeleteWAL(ctx, req.Storage, resp.WALID)
+				if walDeleteErr != nil {
+					var merr *multierror.Error
+					merr = multierror.Append(merr, err)
+					merr = multierror.Append(merr, fmt.Errorf("failed to clean up WAL from failed role creation: %w", walDeleteErr))
+					err = merr.ErrorOrNil()
+				}
+			}
 			return nil, err
 		}
 		// guard against RotationTime not being set or zero-value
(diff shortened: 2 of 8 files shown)


rotation.go
```diff
@@ -72,13 +72,15 @@ func (b *backend) populateQueue(ctx context.Context, s logical.Storage) {
...
 				if err := framework.DeleteWAL(ctx, s, walEntry.walID); err != nil {
 					log.Warn("unable to delete WAL", "error", err, "WAL ID", walEntry.walID)
 				}
 			} else {
-				log.Info("adjusting priority for Role")
+				log.Info("found WAL for role",
+					"role", item.Key,
+					"WAL ID", walEntry.walID)
 				item.Value = walEntry.walID
-				item.Priority = time.Now().Unix()
```
tomhjp, line 81: This change is the main fix

briankassouf: Isn't still a good idea to process WALs immediately, particularly in the case where the a manual rotation has happened but failed in the middle? The manual rotation attempt would not be accounted for in the NextRotationTime() calculation.

jasonodonnell: Agreed with @briankassouf.

tomhjp: That's a good question. Is there any reason the manual rotation endpoint needs to use the queue at all? It would simplify things a lot if we can just return errors on that endpoint and keep the mechanisms for periodic rotation and manual rotation's retries separate.

I do think that should be the long term direction, but it's maybe more problematic for a patch version.

briankassouf: I think the WAL exists to handle errors mid-rotation; which is different than the reason for the queue. Otherwise it's hard to reason about "recovery". I suspect we'll need some kind of WAL approach in either case.

jasonodonnell: The philosophy around using the queue was to standardize how passwords are rotated across all the endpoints. The difference being a manual rotation should be high priority.

path_rotate.go
```diff
@@ -159,9 +161,16 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 		return nil, pushErr
 	}
 
+	if err != nil {
+		return &logical.Response{
```
tomhjp, line 165: I haven't tested yet, but I think this combination of non-error response object, plus an error will return a 400 bad request status code from the API, which isn't ideal.

rotation.go
tomhjp: Thanks for the input. After discussing yesterday, we've taken a fairly different approach to the fix now, instead focussing on making sure we don't allow multiple WALs per role, and therefore trusting that the presence of a WAL implies a manual rotation was requested if it's prior to the scheduled rotation time.

path_rotate.go
tomhjp: Perhaps it would be better to just return what is currently a warning string as an error, and include the underlying error in that string?

tomhjp:
Rebased against `master` to resolve some conflicts

path_rotate.go
briankassouf: Good point, yeah including it in the error message sounds good!

path_static_roles.go
```diff
@@ -233,6 +234,9 @@ func (b *backend) pathStaticRoleCreateUpdate(ctx context.Context, req *logical.R
 		}
```
briankassouf, line 234: If we get back an error here should we attempt to delete the WAL that was created? It'll be best-effort but IMO worth trying.

tomhjp: Yeah that seems like a good idea. In a similar vein, I've held off adding deletion of WALs on role deletion for now just to keep the patch smaller because it's not critical to the core bug, but do you think it's worth adding that too?

path_static_roles.go
```diff
@@ -247,17 +251,18 @@ func (b *backend) pathStaticRoleCreateUpdate(ctx context.Context, req *logical.R
 		// the queue
 
 		//TODO: Add retry logic
-		_, err = b.popFromRotationQueueByKey(name)
+		// We could be tracking a WAL ID for this role, ensure we keep it when
+		// it's re-added to the queue.
```
calvn, line 255: Is this comment still applicable? When we're popping this item from the queue, it should already have the `walID` if that was previously set right?

rotation.go
```diff
@@ -309,57 +315,66 @@ func (b *backend) setStaticAccountPassword(ctx context.Context, s logical.Storag
...
+	var newPassword string
+	if output.WALID != "" {
+		wal, err := b.findStaticWAL(ctx, s, output.WALID)
+		if err != nil {
+			return output, errwrap.Wrapf("error retrieving WAL entry: {{err}}", err)
+		}
+
+		switch {
+		case wal != nil && wal.NewPassword != "":
+			newPassword = wal.NewPassword
```
calvn, line 339: In the case where we want to be updating due to an existing WAL, should we be using `wal.OldPassword`?

rotation.go
```diff
@@ -309,57 +315,66 @@ func (b *backend) setStaticAccountPassword(ctx context.Context, s logical.Storag
...
+		if err != nil {
+			return output, err
+		}
 		output.WALID, err = framework.PutWAL(ctx, s, staticWALKey, &setCredentialsWAL{
 			RoleName:          input.RoleName,
 			Username:          input.Role.StaticAccount.Username,
 			DN:                input.Role.StaticAccount.DN,
 			NewPassword:       newPassword,
-			OldPassword:       oldPassword,
+			OldPassword:       input.Role.StaticAccount.Password,
```
calvn, line 366: We set OldPassword in here and `findStaticWAL`, but don't see where `OldPassword` is referenced/used.

rotation.go
```diff
@@ -299,8 +303,10 @@ type setStaticAccountOutput struct {
 // tasks must be handled outside of this method.
 func (b *backend) setStaticAccountPassword(ctx context.Context, s logical.Storage, input *setStaticAccountInput) (*setStaticAccountOutput, error) {
 	var merr error
+	// Re-use WAL ID if present, otherwise PUT a new WAL
+	output := &setStaticAccountOutput{WALID: input.WALID}
 	if input == nil || input.Role == nil || input.RoleName == "" {
```
kalafut, line 308: Should the `input == nil` check be before the `input.WALID` reference above? Probably will need to split the conditional if so and we still want to return `output` in the error case.

rotation.go
```diff
@@ -309,57 +315,66 @@ func (b *backend) setStaticAccountPassword(ctx context.Context, s logical.Storag
...
-
-	newPassword, err := b.GeneratePassword(ctx, config)
-	if err != nil {
-		return nil, err
+		return output, errors.New("the config is currently unset")
 	}
 
-	oldPassword := input.Role.StaticAccount.Password
-
 	// Take out the backend lock since we are swapping out the connection
```
kalafut, line 326: What connection is this comment referring to?

rotation.go
kalafut: I'm not following this logic. Wouldn't that simply be a no-op, since the current password would be the OldPassword (because the operation failed, hence the WAL being present).

kalafut:
Can we consider renaming the function `pathRotateCredentialsUpdate` to `pathRotateRootCredentialsUpdate`. I'm continually confused, thinking this one is the non-root one since surely "root" would be called out.

path_static_roles.go
tomhjp: That's right. Maybe I've written it weirdly, but I wanted to leave a comment to highlight that it's important we keep the `item` returned from pop, unlike before.

rotation.go
tomhjp: Good point. Technically, with the way the response from this function is handled, we only _need_ to return the WAL ID if it changes during this function, but I was moving it up to the top just to be defensive. I've moved it down a little.

rotation.go
tomhjp: I guess it was referring to the config write? I've deleted this comment, but it's definitely on my todo list to review the locking for the plugin.

rotation.go
tomhjp: The fact we have a WAL means there was either a scheduled or manually triggered rotation. So it's safest to always try to roll forwards based on the intentions we previously recorded in the WAL. If we rolled back to the old password, we would then want to immediately trigger another attempt at rotating.

rotation.go
tomhjp: Yeah I don't think we need to set it at all, because we never need/want to roll back for non-root roles. If we used WALs for root credential rotation, then it would be helpful to have the old password, but unfortunately we don't use WALs for root credential rotation.

calvn (approved):
Couple of smaller follow-up comments, but the overall implementation looks good!

rotation.go
tomhjp: Yeah I think that should be safe

path_static_roles.go
```diff
@@ -146,7 +147,28 @@ func (b *backend) pathStaticRoleDelete(ctx context.Context, req *logical.Request
...
+		}
+		if wal != nil && name == wal.RoleName {
+			b.Logger().Debug("deleting WAL for deleted role", "WAL ID", walID, "role", name)
+			err = framework.DeleteWAL(ctx, req.Storage, walID)
+			if err != nil {
+				b.Logger().Debug("failed to delete WAL for deleted role", "WAL ID", walID, "error", err)
+				merr = multierror.Append(merr, err)
+			}
+		}
+	}
```
briankassouf, line 169: I wonder if we want to do the WAL delete before the role deletion, that way we have an avenue to retry the operation if the WAL cleanup fails?

rotation.go
```diff
@@ -312,54 +312,64 @@ func (b *backend) setStaticAccountPassword(ctx context.Context, s logical.Storag
...
 			LastVaultRotation: input.Role.StaticAccount.LastVaultRotation,
 		})
 		if err != nil {
 			return output, errwrap.Wrapf("error writing WAL entry: {{err}}", err)
 		}
+		b.Logger().Debug("writing WAL", "role", input.RoleName, "WAL ID", output.WALID)
 	}
 
 	// Update the password remotely.
 	if err := b.client.UpdatePassword(config.LDAP, input.Role.StaticAccount.DN, newPassword); err != nil {
```
briankassouf, line 371: Given all the branching above can we verify `len(newPassword) > 0` before updating the PW? It seems fine now, but this ought to future proof the code a bit

briankassouf (approved):
Just a few minor things, otherwise the changes look good

rotation.go
tomhjp: Seems reasonable. In terms of handling that situation, I've regenerated the password and logged an error. That relies on the generate password function never returning an empty string, because we don't check the new password again, but I hope that's a reasonable assumption.

path_static_roles.go
tomhjp: I tried implementing this, but I think I'd prefer to leave WAL cleanup as a best effort, because although it has its own consistency problems, the other order does as well. If we delete WALs before the role and fail, we'll need to push the item back onto the queue, which itself could fail. In practice though, deleting roles and WALs both rely on the same storage layer, so at least these partial failure scenarios should be very rare.

We could address the inconsistency of leaving WALs lying around fairly robustly by keying WALs on a UUID whose lifetime is attached to a role's storage entry instead of keying WALs on the role name. That obviously wouldn't be a backport candidate though. WDYT?

briankassouf: Makes sense, let's leave it as is for now and revisit any improvements in a subsequent PR

PR #36, opened 2022-02-18
Fix panic from nil logger

jasonodonnell:
We aren't passing in a logger, which can cause Vault to panic if it attempts to use a logger in the SDK `ldaputil` code.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,6 +26,9 @@ func NewClient(conf *logical.BackendConfig) *Client {
 	}
 
 	logger := conf.Logger
+
+	// This can only happen in a test where BackendConfig is created
+	// manually without a logger.
 	if logger == nil {
 		logger = hclog.NewNullLogger()
 	}
--- a/client.go
+++ b/client.go
@@ -7,7 +7,6 @@ import (
 	"github.com/go-ldap/ldif"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-plugin-secrets-openldap/client"
-	"github.com/hashicorp/vault/sdk/logical"
 )
 
 type ldapClient interface {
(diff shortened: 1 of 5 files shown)


fairclothjm (approved):

client.go
```diff
@@ -18,10 +19,13 @@ type ldapClient interface {
 	Execute(conf *client.Config, entries []*ldif.Entry, continueOnError bool) (err error)
 }
 
-func NewClient() *Client {
-	return &Client{
+func NewClient(conf *logical.BackendConfig) *Client {
+	client := &Client{
 		ldap: client.New(),
 	}
+	client.ldap.LDAP.Logger = conf.Logger
```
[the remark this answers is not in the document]
jasonodonnell: It's guaranteed but you can misconfigure it in tests, so I'll add a check.

fairclothjm (approved):

client.go
```diff
@@ -18,10 +20,22 @@ type ldapClient interface {
 	Execute(conf *client.Config, entries []*ldif.Entry, continueOnError bool) (err error)
 }
 
-func NewClient() *Client {
-	return &Client{
+func NewClient(conf *logical.BackendConfig) *Client {
```
calvn, line 23: We probably don't need the whole config and could just accept an `hclog.Logger`.

client.go
```diff
@@ -18,10 +20,22 @@ type ldapClient interface {
...
+
+	logger := conf.Logger
+
+	// This can only happen in a test where BackendConfig is created
+	// manually without a logger.
+	if logger == nil {
+		logger = hclog.NewNullLogger()
+	}
+
+	client.ldap.LDAP.Logger = logger
```
calvn, line 36: Would it be cleaner to pass in a logger on the client itself like so `client.New(logger)`?

client/client.go
```diff
@@ -22,14 +22,14 @@ type Config struct {
 
 func New() Client {
 	return Client{
-		ldap: &ldaputil.Client{
+		LDAP: &ldaputil.Client{
 			LDAP: ldaputil.NewLDAP(),
 		},
```
calvn, line 27: `Logger` could be set here, and we could move the nil check within this constructor?

jasonodonnell: Done.

calvn (approved):

PR #40, opened 2022-08-31
Unifies feature set with the Active Directory secrets engine

austingebauer:
## Overview

This PR unifies the feature set offered by this secrets engine with Vault's [Active Directory Secrets Engine](https://github.com/hashicorp/vault-plugin-secrets-ad) (AD). The primary goal is to allow users of the AD secrets engine to eventually migrate to using this secrets engine without major behavior changes or loss of capability.

Note that a significant portion of this PR is code directly copied from the AD secrets engine.

Notable features and improvements include:
- [Service account check-out](https://www.vaultproject.io/docs/secrets/ad#service-account-check-out) for `ad`, `openldap`, and `racf` [schemas](https://www.vaultproject.io/api-docs/secret/openldap#schema)
- Adds the `userdn` and `userattr` configuration parameters to control how user LDAP 
  search is performed for service account check-out and static roles.
- Adds the `upndomain` configuration parameter to allow construction of a userPrincipalName 
  (UPN) string for authentication.

Notable features left out:
- [Lazy rotation](https://www.vaultproject.io/docs/secrets/ad#a-note-on-lazy-rotation) of passwords for roles. The OpenLDAP secrets engine already supports interval and on-demand password rotation for AD users via [static roles](https://www.vaultproject.io/docs/secrets/openldap#static-roles).

## Testing

I've ported tests for service account check-out over from the AD secrets engine. Existing tests for this secrets engine are passing. I also performed extensive manual testing using Vault namespaces and performance standby instances.


backend.go
```diff
@@ -11,6 +11,38 @@ import (
 	"github.com/hashicorp/vault/sdk/queue"
 )
 
+type backend struct {
```
austingebauer, line 14: This backend struct appears new in the diff, but it is actually relocated. Comments below highlight new members introduced by this PR.

backend.go
```diff
@@ -11,6 +11,38 @@ import (
...
+	// background ticker.
+	credRotationQueue *queue.PriorityQueue
+	cancelQueue       context.CancelFunc
+
+	// managedUsers contains the set of LDAP usernames managed by the secrets engine
+	// static role and check-in/check-out systems. It is used to ensure that users
+	// are exclusively managed by one system and not both. Access to managedUsers is
+	// synchronized by the managedUserLock.
+	managedUsers    map[string]struct{}
+	managedUserLock sync.Mutex
```
austingebauer, line 34: This is new code introduced by this PR. See comment for its purpose.

backend.go
```diff
@@ -11,6 +11,38 @@ import (
...
+
+	// roleLocks is used to lock modifications to roles in the queue, to ensure
+	// concurrent requests are not modifying the same role and possibly causing
+	// issues with the priority queue.
+	roleLocks []*locksutil.LockEntry
+
+	// checkOutLocks are used for avoiding races when working with library sets
+	// in the check-in/check-out system.
+	checkOutLocks []*locksutil.LockEntry
```
austingebauer, line 43: This code is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

backend_checkouts_test.go
```diff
@@ -0,0 +1,579 @@
+package openldap
```
austingebauer, line 1: This code is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

backend.go
```diff
@@ -22,42 +54,49 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
...
 			b.pathRotateCredentials(),
-			b.pathConfig(),
-
-			b.pathDynamicRoles(),
+			b.pathSetCheckIn(),
+			b.pathSetManageCheckIn(),
+			b.pathSetCheckOut(),
+			b.pathSetStatus(),
+			b.pathSets(),
+			b.pathListSets(),
```
austingebauer, line 88: This is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

backend_checkouts_test.go
austingebauer, line 1: This code is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad). There are minor modifications to the setup.

checkout_handler.go
```diff
@@ -0,0 +1,188 @@
+package openldap
```
austingebauer, line 1: This code is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

checkout_handler_test.go
```diff
@@ -0,0 +1,154 @@
+package openldap
```
austingebauer, line 1: This code is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad). There are minor modifications to the setup.

client.go
```diff
@@ -10,12 +10,8 @@ import (
 )
 
 type ldapClient interface {
-	Add(conf *client.Config, req *ldap.AddRequest) error
-	Get(conf *client.Config, dn string) (*client.Entry, error)
-	Del(conf *client.Config, req *ldap.DelRequest) error
```
austingebauer, line 15: These interface methods were unused, so I deleted them.

client.go
```diff
@@ -31,54 +27,43 @@ type Client struct {
 	ldap client.Client
 }
 
-func (c *Client) Get(conf *client.Config, dn string) (*client.Entry, error) {
-	filters := map[*client.Field][]string{
-		client.FieldRegistry.ObjectClass: {"*"},
-	}
+// UpdateDNPassword updates the password for the object with the given DN.
+func (c *Client) UpdateDNPassword(conf *client.Config, dn string, newPassword string) error {
```
austingebauer, line 31: `UpdateRootPassword` was the exact same code as updating a password using a DN, so I removed it. Happy to add it back if anyone finds value in the distinction.

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
 		return errors.New("unable to bind due to lack of configured password")
 	}
 
-	if cfg.BindDN == "" {
-		return errors.New("must provide binddn")
-	}
-
-	origErr := conn.Bind(cfg.BindDN, cfg.BindPassword)
-	if origErr == nil {
+	if cfg.UPNDomain != "" {
```
austingebauer, line 126: This is copied from [client.go#L128-L155](https://github.com/hashicorp/vault-plugin-secrets-ad/blob/main/plugin/client/client.go#L128-L155) to maintain compatibility for users migrating from AD secrets to LDAP secrets.

path_checkouts.go
```diff
@@ -0,0 +1,390 @@
+package openldap
```
austingebauer, line 1: This is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

path_checkouts_test.go
```diff
@@ -0,0 +1,30 @@
+package openldap
```
austingebauer, line 1: This is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

path_config.go
```diff
@@ -76,8 +76,22 @@ func (b *backend) configFields() map[string]*framework.FieldSchema {
 }
 
 func (b *backend) configCreateUpdateOperation(ctx context.Context, req *logical.Request, fieldData *framework.FieldData) (*logical.Response, error) {
+	conf, err := readConfig(ctx, req.Storage)
```
austingebauer, line 79: This API didn't respect prior values set for updates. For example, if a user provided some parameter on create and didn't provide the same parameter on update, its prior value (from the create) would not be retained. I fixed this to operate similar to most other Vault APIs.

path_dynamic_creds.go
```diff
@@ -15,6 +16,30 @@ import (
 	"golang.org/x/text/encoding/unicode"
 )
 
+func (b *backend) pathDynamicCredsCreate() []*framework.Path {
```
austingebauer, line 19: This code is not new (relocated).

path_static_roles.go
```diff
@@ -56,58 +63,17 @@ func (b *backend) pathStaticRoles() []*framework.Path {
...
 			HelpDescription: staticRoleHelpDescription,
 		},
-	}
-}
-
-// fieldsForType returns a map of string/FieldSchema items for the given role
-// type. The purpose is to keep the shared fields between dynamic and static
-// roles consistent, and allow for each type to override or provide their own
-// specific fields
-func fieldsForType(roleType string) map[string]*framework.FieldSchema {
```
austingebauer, line 66: This provided little utility, so I removed it. There was only one type being passed into it. I found the code more readable by including the fields directly into the path fields list (provides at-a-glance understanding at the API without having to jump around functions).

backend.go
austingebauer: This is one approach at exclusively managing users in library sets and static roles. A different approach that I thought of involved listing all static roles and library sets from Vault storage every time we manage a new user. It's simpler but has different performance tradeoffs. Happy to discuss.

path_checkouts.go
```diff
@@ -0,0 +1,390 @@
...
+	return []*framework.Path{
+		{
+			Pattern: libraryPrefix + framework.GenericNameRegex("name") + "/check-in$",
+			Fields: map[string]*framework.FieldSchema{
+				"name": {
+					Type:        framework.TypeLowerCaseString,
+					Description: "Name of the set.",
+					Required:    true,
+				},
+				"service_account_names": {
```
austingebauer, line 186: I'm wondering if `service_account_names` still makes sense, since we'll allow library sets to be created for schemas other than AD (e.g., OpenLDAP, RACF). I know that we currently target users using [UserPrincipalName in AD secrets engine](https://github.com/hashicorp/vault-plugin-secrets-ad/blob/main/plugin/util/secrets_client.go#L69). UserPrincipalName is an attribute of an [AD user](https://docs.microsoft.com/en-us/powershell/module/activedirectory/new-aduser?view=windowsserver2022-ps). AD [service accounts](https://docs.microsoft.com/en-us/powershell/module/activedirectory/new-adserviceaccount?view=windowsserver2022-ps) do not have the UserPrincipleName attribute. I know that the definition of user vs. service account is pretty loose in LDAP. I think `usernames` could make sense as an alternative name (esp. since static roles in the LDAP secrets engine have a [`username`](https://www.vaultproject.io/api-docs/secret/openldap#username) field).

Looking for opinions on this.

path_config.go
robmonte: Should this be considered its own bugfix and get backported? Sounds like a past oversight to me maybe.

austingebauer: Ya, that a good consideration. I'm actually going to remove other isolated bug fixes in this PR to new PRs so that they can be backported.

path_checkout_sets.go
```diff
@@ -0,0 +1,411 @@
+package openldap
```
austingebauer, line 1: This is copied from [vault-plugin-secrets-ad](https://github.com/hashicorp/vault-plugin-secrets-ad).

path_config.go
austingebauer: I've split this out into a separate bug fix PR in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/43. I'll remove the change from this PR as well. Thanks, @robmonte!

path_checkout_sets.go
```diff
@@ -0,0 +1,411 @@
...
+	for _, serviceAccountName := range set.ServiceAccountNames {
+		if err := b.DeleteCheckout(ctx, req.Storage, serviceAccountName); err != nil {
+			return nil, err
+		}
+	}
+	if err := req.Storage.Delete(ctx, libraryPrefix+setName); err != nil {
+		return nil, err
+	}
+
+	b.deleteManagedUsers(set.ServiceAccountNames...)
```
austingebauer, line 364: Missing lock

path_checkouts.go
austingebauer: TODO: Alias of usernames and fallback to service_account_names

path_static_roles.go
```diff
@@ -168,6 +168,8 @@ func (b *backend) pathStaticRoleDelete(ctx context.Context, req *logical.Request
 		}
 	}
 
+	b.deleteManagedUsers(role.StaticAccount.Username)
```
austingebauer, line 171: TODO: lock managed users

path_checkout_sets.go
austingebauer: Done in e821b9d.

path_static_roles.go
austingebauer: Done in e821b9d.

path_checkouts.go
austingebauer: Thinking on this some more -- perhaps "service account" does make sense in the context of check-in/check-out. It feels a little odd to think of checking out a "username", even if under the hood, we're managing password rotation for a user-like object in LDAP.

path_checkouts.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	// Check out the first service account available.
+	for _, serviceAccountName := range set.ServiceAccountNames {
+		if err := b.CheckOut(ctx, req.Storage, serviceAccountName, newCheckOut); err != nil {
+			if err == errCheckedOut {
+				continue
+			}
+			return nil, err
+		}
```
robmonte, line 88: I noticed this doesn't also continue the loop if err is `errNotFound`. I'm thinking this might be because there shouldn't be a case where a name in the Library Set contains a service account name that isn't managed by the plugin. Is that logic correct or on the right track?

path_checkouts.go
robmonte: If the concern here is that this conflates the term "service account" used elsewhere, could just the word "accounts" be a viable middle-ground here? Though it sounds pretty generic, but figured I'd throw it out there.

path_static_roles.go
```diff
@@ -77,10 +81,6 @@ func fieldsForType(roleType string) map[string]*framework.FieldSchema {
 			Type:        framework.TypeString,
 			Description: "The distinguished name of the entry to manage.",
 		},
-		"ttl": {
-			Type:        framework.TypeDurationSecond,
-			Description: "The time-to-live for the password.",
-		},
```
robmonte, line 83: I'm trying to understand why ttl is removed and the static creds seem to be based off of the RotationPeriod field, while dynamic creds are instead still using ttl? Perhaps I'm missing something bigger picture.

robmonte (commented):
Things look good to me, mostly just have a few questions for my own understanding.

path_checkouts.go
austingebauer: Yep, that's correct! It shouldn't be possible to hit that case. If it happens, there would be a bug in the code somewhere.

path_static_roles.go
austingebauer: If you look at the usage of `fieldsForType`, it is only called from `pathStaticRoles`. Since static roles don't use the `ttl` field, I chose to remove it from this list. Dynamic creds use the [default_ttl and max_ttl fields](https://github.com/hashicorp/vault-plugin-secrets-openldap/blob/main/path_dynamic_roles.go#L54-L61).

path_static_roles.go
austingebauer: I decided to add this back to minimize the diff in this PR.

path_static_roles.go
austingebauer: I had actually removed `fieldsForType` earlier in this PR (see my comment in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40#discussion_r960103063) because it doesn't provide much value.

path_dynamic_creds.go
```diff
@@ -15,6 +16,30 @@ import (
...
+	return []*framework.Path{
```
swenson, line 20: Is this function named correctly still? It seems to only have the read operation defined.

swenson (commented):
Looks pretty good. A few minor comments.

path_static_roles.go
robmonte: Ahh I see. I hadn't actually caught on yet that static roles don't have TTLs in general. I figured it only applied to the password in the case of static roles but rotation already covers that. That's good to have learned!

path_checkouts_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func TestCheckInAuthorized(t *testing.T) {
```
tomhjp, line 9: Might be a nice candidate for a table driven test, which reduces duplication, and separates data from logic so it's easier to read the test case data.

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
+	if cfg.BindDN != "" {
+		origErr := conn.Bind(cfg.BindDN, cfg.BindPassword)
+		if origErr == nil {
+			return nil
+		}
+		if !shouldTryLastPwd(cfg.LastBindPassword, cfg.LastBindPasswordRotation) {
+			return origErr
+		}
+		if err := conn.Bind(cfg.BindDN, cfg.LastBindPassword); err != nil {
+			// Return the original error because it'll be more helpful for debugging.
```
tomhjp, line 150: I agree in most cases it is likely to be the most helpful error, but could we return both with some added context for maximum utility? e.g. "Attempted with current and previous password and got the following errors..."

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
-	return conn.Bind(cfg.BindDN, cfg.LastBindPassword)
+	if cfg.BindDN != "" {
+		origErr := conn.Bind(cfg.BindDN, cfg.BindPassword)
+		if origErr == nil {
+			return nil
+		}
+		if !shouldTryLastPwd(cfg.LastBindPassword, cfg.LastBindPasswordRotation) {
+			return origErr
+		}
+		if err := conn.Bind(cfg.BindDN, cfg.LastBindPassword); err != nil {
```
tomhjp, line 149: I think if this call succeeds, we'll fall through to the `"must provide binddn or upndomain"` error. Using a switch-case (or an if-else chain, but then I think lots of Go linters will complain about unnecessary indentation) instead of if statements would make it clearer that we expect to take precisely 1 of the 3 code paths, and easier to avoid that bug.

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
+		origErr := conn.Bind(fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain), cfg.BindPassword)
+		if origErr == nil {
+			return nil
+		}
+		if !shouldTryLastPwd(cfg.LastBindPassword, cfg.LastBindPasswordRotation) {
+			return origErr
+		}
+		if err := conn.Bind(fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain), cfg.LastBindPassword); err != nil {
+			// Return the original error because it'll be more helpful for debugging.
+			return origErr
```
tomhjp, line 136: As below, if we're content with changing the error to be the original error, maybe worth going all in and returning both errors.

path_static_roles.go
```diff
@@ -39,10 +39,14 @@ func (b *backend) pathStaticRoles() []*framework.Path {
 			ExistenceCheck: b.pathStaticRoleExistenceCheck,
 			Operations: map[logical.Operation]framework.OperationHandler{
 				logical.UpdateOperation: &framework.PathOperation{
-					Callback: b.pathStaticRoleCreateUpdate,
+					Callback:                    b.pathStaticRoleCreateUpdate,
+					ForwardPerformanceStandby:   true,
+					ForwardPerformanceSecondary: true,
```
tomhjp, line 44: Is this a bug fix or an optimisation?

path_dynamic_creds.go
```diff
@@ -15,6 +16,30 @@ import (
...
+		{
+			Pattern: path.Join(dynamicCredPath, framework.MatchAllRegex("name")),
```
tomhjp, line 22: Is `MatchAllRegex` definitely correct here? I'm not sure if more is happening under the hood, but the regex looks like it allows slashes too. I know this is pre-existing though, so no need to actually address in this PR.

path_dynamic_creds.go
```diff
@@ -15,6 +16,30 @@ import (
...
+			Operations: map[logical.Operation]framework.OperationHandler{
+				logical.ReadOperation: &framework.PathOperation{
+					Callback: b.pathDynamicCredsRead,
+				},
+			},
+			HelpSynopsis: "Request LDAP credentials for a dynamic role. These credentials are " +
+				"created within the LDAP system when querying this endpoint.",
+			HelpDescription: "This path requests new LDAP credentials for a certain dynamic role. " +
+				"The credentials are created within the LDAP system based on the creation_ldif " +
+				"specified within the dynamic role configuration.",
```
tomhjp, line 38: Maybe worth adding a note about eventual consistency?

tomhjp (commented):
I've still got path_checkout*.go and some tests to go but I'll checkpoint my review here briefly.

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
-	origErr := conn.Bind(cfg.BindDN, cfg.BindPassword)
-	if origErr == nil {
+	if cfg.UPNDomain != "" {
+		origErr := conn.Bind(fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain), cfg.BindPassword)
+		if origErr == nil {
+			return nil
+		}
+		if !shouldTryLastPwd(cfg.LastBindPassword, cfg.LastBindPasswordRotation) {
+			return origErr
+		}
```
maxcoulombe, line 133: I think errors should be wrapped and not skip levels when traversing between different systems. It makes it much easier to debug what exactly happened from a log file. I also think errors should never be swallowed, the code should either handle them or bubble the information up.

`return fmt.Errorf("unable to bind new password: %w", origErr)`

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
+		if err := conn.Bind(fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain), cfg.LastBindPassword); err != nil {
+			// Return the original error because it'll be more helpful for debugging.
+			return origErr
+		}
```
maxcoulombe, line 137: Same here, I feel like obfuscating the original error could make bugs and problem much harder to troubleshoot.

`return fmt.Errorf("unable to bind last password following problem when binding new password: %w: %w", err, origErr)`

path_static_roles.go
```diff
@@ -213,18 +217,31 @@ func (b *backend) pathStaticRoleCreateUpdate(ctx context.Context, req *logical.R
...
-	if username == "" {
+	if isCreate && username == "" {
 		return logical.ErrorResponse("username is a required field to manage a static account"), nil
 	}
+	if isCreate && b.isManagedUser(username) {
+		return logical.ErrorResponse("%q is already managed by the secrets engine", username), nil
+	}
+	if !isCreate && username != role.StaticAccount.Username {
+		return logical.ErrorResponse("cannot update static account username"), nil
+	}
```
maxcoulombe, line 234: Seems like there is a bit of test coverage missing for these conditions?

maxcoulombe (approved):
Nice job that's super clean; my comments are mostly styling and not logic as I'm not super familiar with LDAP so I'm approving the PR but if you have questions don't hesitate!

tomhjp (approved):
LGTM. I think there's 1 trivial bug from my previous comments, and otherwise just some open questions and nits. I'll try to give this a test drive soon as well.

path_dynamic_creds.go
austingebauer: I _think_ it makes sense. The read operation in this case creates an LDAP user and secret in Vault ([path_dynamic_creds.go#L18-L92](https://github.com/hashicorp/vault-plugin-secrets-openldap/blob/main/path_dynamic_creds.go#L18-L92)). One of those Vault-y things where a read creates something 🙂

client/client.go
```diff
@@ -122,25 +123,41 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
 	}
 
-	if cfg.BindDN == "" {
-		return errors.New("must provide binddn")
-	}
-
-	origErr := conn.Bind(cfg.BindDN, cfg.BindPassword)
-	if origErr == nil {
+	if cfg.UPNDomain != "" {
+		origErr := conn.Bind(fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain), cfg.BindPassword)
```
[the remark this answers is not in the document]
austingebauer: I would've expected that we escape the `UPNDomain` as well. I'm hesitant to change the behavior at this point because we don't escape it in both LDAP auth ([client.go#L140-L178](https://github.com/hashicorp/vault/blob/main/sdk/helper/ldaputil/client.go#L140-L178)) and AD secrets engine ([client.go#L128-L141](https://github.com/hashicorp/vault-plugin-secrets-ad/blob/main/plugin/client/client.go#L128-L141)).

backend.go
```diff
@@ -105,17 +118,101 @@ type backend struct {
...
+	if err != nil {
+		return err
+	}
+	for _, roleName := range staticRoles {
+		staticRole, err := b.staticRole(ctx, s, roleName)
+		if err != nil {
+			return err
+		}
+		if staticRole == nil {
+			continue
```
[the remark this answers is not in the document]
austingebauer: It's mostly a defensive check in the sense that we can't do much with a nil `staticRole` here. Like you've mentioned, it's not possible to have that race occur where this is being called in the plugin lifecycle. It'd be good to know if somehow this did happen (could indicate a bug in the plugin lifecycle where requests are processed before the backend InitializeFunc returns). I'm going to add a debug log. I'll consider holding the backend lock, but I first need to understand it's usage throughout this plugin better.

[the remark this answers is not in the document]
austingebauer: I've added a comment and debug log in 47e46f9.

backend.go
```diff
@@ -105,17 +118,101 @@ type backend struct {
...
+	if err != nil {
+		return err
+	}
+	for _, setName := range librarySets {
+		set, err := readSet(ctx, s, setName)
+		if err != nil {
+			return err
+		}
+		if set == nil {
+			continue
```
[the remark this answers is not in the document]
austingebauer: I've added a comment and debug log in https://github.com/hashicorp/vault-plugin-secrets-openldap/commit/47e46f9cd3bfc458266cc76c69b7cc9ce75a49b5.

checkout_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	if config == nil {
+		return errors.New("the config is currently unset")
+	}
+
+	newPassword, err := b.GeneratePassword(ctx, config)
+	if err != nil {
+		return err
+	}
+	if err := b.client.UpdateUserPassword(config.LDAP, serviceAccountName, newPassword); err != nil {
```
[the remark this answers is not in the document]
austingebauer: Yep, WALs are used for the static role rotation mechanism. There weren't any WALs in the library set code coming from AD secrets. I will evaluate if this is something we _should_ do and create a task if so. For this PR, I won't be adding a WAL and rollback mechanism for library sets.

backend.go
```diff
@@ -105,17 +118,101 @@ type backend struct {
...
+	}
+}
+
+// deleteManagedUsers sets the user as unmanaged by the secrets engine.
+// Must be called with the managedUserLock held.
+func (b *backend) deleteManagedUsers(users ...string) {
+	for _, user := range users {
+		delete(b.managedUsers, user)
+	}
 }
```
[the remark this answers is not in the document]
austingebauer: I agree! I tried a few different things here. If these functions handle the locking themselves, there are races that can occur. It's likely best to inline them. I'll take a second look at this and let you know what I come up with.

client.go
```diff
@@ -3,14 +3,15 @@ package openldap
...
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-plugin-secrets-openldap/client"
 )
 
 type ldapClient interface {
-	UpdatePassword(conf *client.Config, dn string, newPassword string) error
-	UpdateRootPassword(conf *client.Config, newPassword string) error
+	UpdateDNPassword(conf *client.Config, dn string, newPassword string) error
+	UpdateUserPassword(conf *client.Config, user, newPassword string) (err error)
 	Execute(conf *client.Config, entries []*ldif.Entry, continueOnError bool) (err error)
```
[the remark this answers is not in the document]
austingebauer: Whoops! Changed to unnamed in 7cf2bbc.

path_checkouts_test.go
austingebauer: Indeed! I also prefer the table driven tests. These are copied from AD secrets engine, and I left them unmodified to keep the diff small. I can create a task to improve these tests after this PR is merged.

backend.go
```diff
@@ -105,17 +118,101 @@ type backend struct {
...
+	return exists
+}
+
+// setManagedUsers sets the given users as managed by the secrets engine.
+// Must be called with the managedUserLock held.
+func (b *backend) setManagedUsers(users ...string) {
+	for _, user := range users {
+		b.managedUsers[user] = struct{}{}
+	}
+}
```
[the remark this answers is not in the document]
austingebauer: Good call 👍 Changed in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/124c5fdb814c87003d94921bce4f74247a8948d8.

client.go
```diff
@@ -3,14 +3,15 @@ package openldap
...
 	"github.com/go-ldap/ldif"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-plugin-secrets-openldap/client"
 )
 
 type ldapClient interface {
-	UpdatePassword(conf *client.Config, dn string, newPassword string) error
-	UpdateRootPassword(conf *client.Config, newPassword string) error
+	UpdateDNPassword(conf *client.Config, dn string, newPassword string) error
+	UpdateUserPassword(conf *client.Config, user, newPassword string) (err error)
```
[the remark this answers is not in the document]
austingebauer: No specific reason. I just didn't notice the inconsistency. I changed this to unnamed in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/7cf2bbc812410b4e52ed4383dcd350d6b7d4492e.

path_static_roles.go
austingebauer: It's just an optimization.

path_dynamic_creds.go
austingebauer: Good find, it wasn't correct. I've updated this to use `framework.GenericNameRegex` in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/b1426f2ee4809afff541a7d27ed2fd872ade1894.

backend.go
```diff
@@ -22,51 +22,64 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
...
 	// Create a context with a cancel method for processing any WAL entries and
 	// populating the queue
 	ictx, cancel := context.WithCancel(context.Background())
 	b.cancelQueue = cancel
 
-	// Load queue and kickoff new periodic ticker
+	// Load static role queue and kickoff new periodic ticker
 	go b.initQueue(ictx, initRequest)
 
 	return nil
```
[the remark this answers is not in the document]
austingebauer: Done in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/4cee58232c561808dc4810e602fd14d273fafa35.

path_static_roles.go
austingebauer: Good catch -- will add coverage for this!

client/client.go
austingebauer: I agree. I didn't modify this much to preserve the [logic](https://github.com/hashicorp/vault-plugin-secrets-ad/blob/main/plugin/client/client.go#L128-L156) from AD secrets, but I think it's a worthwhile improvement. I'm going to update this section of the code to incorporate your and Tom's comments. Will provide the commit for further feedback when done.

path_checkout_sets.go
```diff
@@ -0,0 +1,413 @@
...
+			},
+			ExistenceCheck:  b.operationSetExistenceCheck,
+			HelpSynopsis:    setHelpSynopsis,
+			HelpDescription: setHelpDescription,
+		},
+	}
+}
+
+func (b *backend) operationSetExistenceCheck(ctx context.Context, req *logical.Request, fieldData *framework.FieldData) (bool, error) {
+	set, err := readSet(ctx, req.Storage, fieldData.Get("name").(string))
```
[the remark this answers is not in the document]
austingebauer: It will return `false, nil` if name is empty. Just double checked this to be sure.

path_checkouts.go
austingebauer: That's a good suggestion as well. I'm going to leave this as "service account" at this point.

path_dynamic_creds.go
austingebauer: Added a note in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/8911ba44f573b226dff33f490ca13f016054686f 👍

path_checkout_sets.go
```diff
@@ -0,0 +1,413 @@
...
+	ServiceAccountNames       []string      `json:"service_account_names"`
+	TTL                       time.Duration `json:"ttl"`
+	MaxTTL                    time.Duration `json:"max_ttl"`
+	DisableCheckInEnforcement bool          `json:"disable_check_in_enforcement"`
+}
+
+// Validate ensures that a set meets our code assumptions that TTLs are set in
+// a way that makes sense, and that there's at least one service account.
+func (l *librarySet) Validate() error {
+	if len(l.ServiceAccountNames) < 1 {
```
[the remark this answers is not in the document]
austingebauer: Good catch. It's possible to have an empty `service_account_names` value when given like: `service_account_names="user1,"`. Internal server error doesn't make sense, so I'll add validation earlier as you've suggested. Thanks!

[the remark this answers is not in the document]
austingebauer: I added validation to catch empty an service account name and test coverage for the `Validate()` in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/336e40ad61ae29814eed22f4fa6de2451edf2f89.

path_checkout_sets.go
```diff
@@ -0,0 +1,413 @@
...
+				},
+			},
+			HelpSynopsis:    pathListSetsHelpSyn,
+			HelpDescription: pathListSetsHelpDesc,
+		},
+	}
+}
+
+func (b *backend) setListOperation(ctx context.Context, req *logical.Request, _ *framework.FieldData) (*logical.Response, error) {
+	keys, err := req.Storage.List(ctx, libraryPrefix)
```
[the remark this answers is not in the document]
austingebauer: It could be inconsistent, but I can't think of any significant consequences. I haven't seen synchronization involved in list operations like this. It will return the list from storage at that point in time.

client/client.go
austingebauer: This bug also exists in AD secrets ([client.go#L143-L156](https://github.com/hashicorp/vault-plugin-secrets-ad/blob/main/plugin/client/client.go#L143-L156)). I'll get it fixed up in both places.

austingebauer: Looking closer at AD, this bug (fixed here in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/42) also exists. Will also get that fixed in AD secrets.

client/client.go
austingebauer: Good idea. Refactored this in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/d7d27a848321e1d1fd77395453c338546715a0fa. Both errors are now included along with a contextual prefix similar to your suggestion here.

client/client.go
austingebauer: Done in https://github.com/hashicorp/vault-plugin-secrets-openldap/commit/d7d27a848321e1d1fd77395453c338546715a0fa. More context in the other related comment.

client/client.go
austingebauer: I've refactored this to fix the bug in https://github.com/hashicorp/vault-plugin-secrets-openldap/commit/d7d27a848321e1d1fd77395453c338546715a0fa. Thanks!

path_static_roles.go
austingebauer: Added coverage for this in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/60d24c919929149777d6f6c264c7682a8c7dbc65. Thanks!

backend.go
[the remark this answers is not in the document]
austingebauer: I ended up removing these functions and inlining the behavior in the operation handlers. You can see the diff in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40/commits/55a59b773296886db499bd48024e9934c257b632. 

I'm now wondering if it's even less obvious that the lock should be held when accessing `b.managedUsers` 🤔 The functions were indeed brittle, but their existence provided an additional opportunity to document that the `managedUserLock` must be held. I'm good either way on this one. Let me know if you have thoughts now that you can see the diff. 

Thanks again!

austingebauer:
I've addressed all of the feedback here. I'm going to give this another quick test tomorrow morning. If all is good, I'll be merging this.

client/client.go
```diff
@@ -122,25 +123,46 @@ func bind(cfg *Config, conn ldaputil.Connection) error {
...
+	// Determine the user to bind with
+	var bindUser string
+	switch {
+	case cfg.UPNDomain != "":
+		bindUser = fmt.Sprintf("%s@%s", ldaputil.EscapeLDAPValue(cfg.BindDN), cfg.UPNDomain)
+	case cfg.BindDN != "":
+		bindUser = cfg.BindDN
+	default:
+		return errors.New("must provide binddn or upndomain")
 	}
```
tomhjp, line 135: Nice! This is so much better, I hadn't noticed the fact this was the only difference between the code paths anyway.

tomhjp (approved):
I can see your point about documentation for locking managed users, but I think it's clear enough as it is. I have no objections either way though.

client/client.go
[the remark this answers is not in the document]
swenson: :+1:

PR #43, opened 2022-09-07
Fixes config updates to retain prior values set

austingebauer:
## Overview

This PR fixes config updates so that they retain prior values in Vault storage. For example, if a user provided some parameter on create and didn't provide the same parameter on update, its prior value (from the create) would not be retained. This makes the config update API semantics similar to most other Vault APIs.

## Testing

I've added a test case to this PR to ensure prior values are retained after an update.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -76,8 +76,23 @@ func (b *backend) configFields() map[string]*framework.FieldSchema {
 }
 
 func (b *backend) configCreateUpdateOperation(ctx context.Context, req *logical.Request, fieldData *framework.FieldData) (*logical.Response, error) {
+	conf, err := readConfig(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+	if conf == nil {
+		conf = new(config)
+		conf.LDAP = new(client.Config)
+	}
+
+	// Use the existing ldap client config if it is set
+	var existing *ldaputil.ConfigEntry
+	if conf.LDAP != nil && conf.LDAP.ConfigEntry != nil {
+		existing = conf.LDAP.ConfigEntry
+	}
+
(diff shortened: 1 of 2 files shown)


path_config.go
```diff
@@ -76,8 +76,23 @@ func (b *backend) configFields() map[string]*framework.FieldSchema {
 }
 
 func (b *backend) configCreateUpdateOperation(ctx context.Context, req *logical.Request, fieldData *framework.FieldData) (*logical.Response, error) {
```
maxcoulombe, line 78: Is there a difference between PUT/POST at the API level?

Right now a create operation would retain existing values if the plugin config already exists; whereas as a user when I call explicitly a create operation I expect APIs to either throw an error or set the config as it was defined without keeping values implicitly depending on if overriding is allowed.

If create/update use the same commands/api calls then this comment doesn't apply since users can't choose between the 2 operations.

path_config_test.go
```diff
@@ -228,6 +228,68 @@ func TestConfig_Update(t *testing.T) {
 			t.Fatalf("no response expected, got: %#v", resp)
 		}
 	})
+
+	t.Run("update retains prior config values in storage", func(t *testing.T) {
```
maxcoulombe, line 232: Should we add a test to document the expected behaviour for create operations on an existing config?

maxcoulombe (approved):

path_config_test.go
austingebauer: Similar to above, it's actually not possible to hit this scenario since we don't define an `ExistenceCheck` here. The backend will always receive a `logical.UpdateOperation` (for both HTTP POST and PUT). Good idea if they were treated as different operations though 👍

PR #44, opened 2022-09-08
Adds the last_password field to the static role credential response

austingebauer:
## Overview

This PR adds the `last_password` field to the static role [credentials response](https://www.vaultproject.io/api-docs/secret/openldap#static-role-passwords). This is carried over from the AD secrets engine role [credential response](https://www.vaultproject.io/api-docs/secret/ad#sample-get-response).

## Testing

I've added test coverage for `last_password` after a static role rotation in this PR.


go.mod
```diff
@@ -28,6 +27,7 @@ require (
 	github.com/go-asn1-ber/asn1-ber v1.5.1 // indirect
 	github.com/golang/protobuf v1.4.2 // indirect
 	github.com/golang/snappy v0.0.1 // indirect
+	github.com/hashicorp/errwrap v1.1.0 // indirect
```
vinay-gopalan, line 30: Looks like we removed the sole spot this package was used. Is it safe to no longer include this package in `go.mod` or are there build/consistency reasons for which we should keep them around?

path_static_roles_test.go
```diff
@@ -78,116 +78,6 @@ func TestRoles(t *testing.T) {
...
-		if resp.Data["rotation_period"] != float64(5) {
-			t.Fatalf("expected rotation_period to be %d but got %s", 5, resp.Data["rotation_period"])
-		}
-
-		if resp.Data["last_vault_rotation"] == nil {
-			t.Fatal("expected last_vault_rotation to not be empty")
-		}
-	})
-
-	t.Run("missing dn", func(t *testing.T) {
```
vinay-gopalan, line 151: Why are we removing the `missing dn` test? I see that the `happy path` test has been duplicated so that one makes sense, but wasn't clear on this one :)

vinay-gopalan (commented):
Looks good! Had a couple questions :)

path_static_roles_test.go
austingebauer: Whoops! Went a little too far when removing the duplicate test. Good catch, thanks! Added back in 1572a37.

go.mod
austingebauer: This is an indirect/transitive dependency. `go mod tidy` doesn't yield a diff. 

It comes from go-multierror:
```sh
go mod why github.com/hashicorp/errwrap

# github.com/hashicorp/errwrap
github.com/hashicorp/vault-plugin-secrets-openldap
github.com/hashicorp/go-multierror
github.com/hashicorp/errwrap
```

vinay-gopalan: Ah gotcha, thanks for confirming!

vinay-gopalan (approved):

PR #46, opened 2022-09-09
Adds userattr config parameter with default based on schema

austingebauer:
## Overview

This PR adds the `userattr` parameter to the config endpoint. Its default will be set based on the `schema` value set in the config. The `userattr` will be used to target specific LDAP user attributes in the upcoming check-in/check-out system introduced in https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/40. In this PR, it is not used for any functionality outside of being set in the config.

## Testing

I've added test cases to this PR to cover the default behavior based on the `schema` value.


vinay-gopalan (approved):
LGTM!

PR #57, opened 2023-03-16
Fixes decoding of prior WAL entries with missing password_policy field

austingebauer:
## Overview

This PR fixes a potential panic introduced by https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/56. The problem is that WAL entries that exist prior to a plugin upgrade might contain missing key/value pairs that the code expected to find in the map. 

To make this more robust to future changes to the WAL data, I've changed it to use mapstructure. This is a common pattern (example in [database/rollback.go](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/rollback.go#L38-L42)).

The test that I wrote panics _before_ the changes introduced by this PR.

## Related Issues/Pull Requests
- https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/56

## Contributor Checklist

- [x] Backwards compatible

### DIFF
--- a/rotation.go
+++ b/rotation.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/vault/sdk/helper/locksutil"
 	"github.com/hashicorp/vault/sdk/logical"
 	"github.com/hashicorp/vault/sdk/queue"
+	"github.com/mitchellh/mapstructure"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


rotation.go
```diff
@@ -256,21 +257,21 @@ func (b *backend) findStaticWAL(ctx context.Context, s logical.Storage, id strin
...
-		Username:       data["username"].(string),
-		DN:             data["dn"].(string),
-		PasswordPolicy: data["password_policy"].(string),
-	}
-	lvr, err := time.Parse(time.RFC3339, data["last_vault_rotation"].(string))
+		walID:        id,
+		walCreatedAt: wal.CreatedAt,
+	}
+	d, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+		DecodeHook: mapstructure.StringToTimeHookFunc(time.RFC3339),
```
robmonte, line 265: This is needed because the missing entry you mentioned is the PasswordPolicy, which was added in the previous bugfix right?

Also, I haven't seen a custom DecoderConfig used like this until now. Am I interpreting the API docs for DecodeHook correctly - decode will be applied to every value in the struct, but `StringToTimeHookFunc` checks if a string is a parseable `time.Time` value and does nothing if it isn't but converts the time if it is?

austingebauer: Yep, that's right! I followed the same pattern in the code which was not checking the WAL data (a map) for existence of each key (see [rotation.go#L263-L267](https://github.com/hashicorp/vault-plugin-secrets-openldap/blob/main/rotation.go#L263-L267)). This is what caused a panic when "password_policy" was empty.

That's correct. If the source field is a string type and the target field is a time.Time type, `StringToTimeHookFunc` will convert the string to a time.Time with the given format (`time.RFC3339`).

fairclothjm (approved):

robmonte (approved):
LGTM!

PR #75, opened 2023-08-10
prevent overwriting of schema and password_policy values on update of…

elliesterner:
`password_policy` and `schema` were getting overwritten on update. 

Example:
```
# create config
vault write open/config \                                                                   
  binddn="cn=admin,dc=hashicorp,dc=com" \
  bindpass="admin" schema="ad" password_policy="3" \
  url="ldap://ldap"

# check that password_policy and schema are there
vault read open/config

# update config
vault write open/config \ 
  bindpass="admin2"

# see that password_policy is gone and schema defaults to "openldap" and not "ad"
vault read open/config
```

# Overview
A high level description of the contribution, including:
Who the change affects or is for (stakeholders)?
Anyone who updates their ldap config

What is the change?
Check to see if there is an existing value in storage for `password_policy` and `schema`. If the request does not update these properties, maintain the original values. 

Why is the change needed?
Values were getting deleted or set back to default.

How does this change affect the user experience (if at all)?

# Contributor Checklist
- [ ] Backwards compatible

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -136,6 +136,13 @@ func (b *backend) configCreateUpdateOperation(ctx context.Context, req *logical.
 	passLength := rawPassLength.(int)
 
 	schema := fieldData.Get("schema").(string)
+	_, schemaChanged := fieldData.Raw["schema"]
+
+	// if update operation and schema not updated in raw payload, keep existing schema
+	if existing != nil && !schemaChanged {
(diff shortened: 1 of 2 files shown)


jasonodonnell (approved):

fairclothjm (approved):
LGTM, thanks!

ltcarbonell (approved):

PR #78, opened 2023-09-01
Prepare for v0.11.2 release

fairclothjm:
Fix broken import `github.com/armon/go-metrics` -> `github.com/hashicorp/go-metrics` which caused auto dep upgrade job to [fail](https://github.com/hashicorp/vault-plugin-secrets-openldap/actions/runs/6043648890/job/16401017141).

Also, update everything else while we are here


austingebauer (approved):
Looks good when test issue resolved 👍

PR #83, opened 2023-10-24
add configuration parameters to skip password rotation on import

kpcraig:
This adds a new parameter to both the ldap configuration and the static role that will skip the initial password rotation on import. While this means vault won't know the password for the static user (/ldap/static-cred/role-name won't return a password), it does mean that an admin can pre-load users into ldap (say, during an AD->LDAP migration), without immediately changing the passwords.

This setting is off by default to retain backwards compatibility.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -93,6 +93,10 @@ func (b *backend) configFields() map[string]*framework.FieldSchema {
 		Type:        framework.TypeString,
 		Description: "Password policy to use to generate passwords",
 	}
+	fields["skip_static_role_import_rotation"] = &framework.FieldSchema{
+		Type:        framework.TypeBool,
+		Description: "Whether to skip the 'import' rotation.",
+	}
 
 	// Deprecated
 	fields["length"] = &framework.FieldSchema{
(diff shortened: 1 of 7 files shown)


path_config.go
```diff
@@ -267,8 +278,9 @@ func (b *backend) configDeleteOperation(ctx context.Context, req *logical.Reques
 }
 
 type config struct {
-	LDAP           *client.Config
-	PasswordPolicy string `json:"password_policy,omitempty"`
+	LDAP                         *client.Config
+	PasswordPolicy               string `json:"password_policy,omitempty"`
+	SkipStaticRoleImportRotation bool   `json:"skip_static_role_import_rotation"`
```
raymonstah, line 283: It would get lengthy, but I wonder if adding some notion of `password` would make this more obvious.
`skip_static_role_import_password_rotation`. Maybe it's redundant.. thoughts?

path_static_roles.go
```diff
@@ -280,14 +285,49 @@ func (b *backend) pathStaticRoleCreateUpdate(ctx context.Context, req *logical.R
...
+				return nil, err
+			}
+
+			// set the item
+			item = &queue.Item{
+				Key: name,
+			}
+			// synthetically set lvr to now, so that it gets queued correctly
+			lvr = time.Now()
+			break
```
raymonstah, line 329: To future proof ourselves from accidentally skipping over additional logic that gets added in this `case` block, can we use `else` here?

path_config.go
kpcraig: i don't mind adding it if people want it - i think rotation implies password in Vault contexts, so i don't think we _need_ it

raymonstah (approved):

fairclothjm:
Should we add a changelog entry for this?

PR #85, opened 2023-10-31
add changelog for new ldap parameter

kpcraig:
I had made this change, and then forgot to commit/push it somehow??


kpcraig:
i know there are a lot of extra commits here because i was lazy about the branch, but since we're squashing it shouldn't matter beyond what is actually in the diff

austingebauer (approved):

PR #91, opened 2024-02-13
bug fix for rotate password when upndomain is not set

raymonstah:
# Overview
Attempts to fix https://github.com/hashicorp/vault/issues/25249

Also adds a couple of dockertests to ensure the rotate password functionality works as expected.


fairclothjm (approved):

austingebauer (approved):
👍

PR #99, opened 2024-03-27
Don't leak timers

swenson:
When in the rollback loop, a timer was created, but if the context is finished, then the timer's channel would leak.

In addition, the function said it was doing exponential backoff (presumably from 0 to 512 seconds), but the arguments to the `Pow()` function were reversed, so it was doing quadratic backoff from 0 to 81 seconds.

Also, there was no test for the rollback function, so one was added.


path_rotate.go
```diff
@@ -181,26 +187,32 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
...
+		timer := time.NewTimer(nextsleep)
 		select {
 		case <-timer.C:
 		case <-ctx.Done():
+			if !timer.Stop() {
+				<-timer.C // drain the channel so that it will be garbage collected
+			}
 			// Outer environment is closing.
 			return fmt.Errorf("unable to roll back password because enclosing environment is shutting down")
 		}
```
benashz, line 210: nit:
```suggestion
			return fmt.Errorf("unable to rollback password because enclosing environment is shutting down")
```

path_rotate.go
```diff
@@ -7,14 +7,20 @@ import (
...
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/backoff"
 	"github.com/hashicorp/vault/sdk/logical"
 	"github.com/hashicorp/vault/sdk/queue"
 )
 
+var (
+	rollbackAttempts    = 10
+	minRollbackDuration = 1 * time.Second
+	maxRollbackDuration = 100 * time.Second
```
vinay-gopalan, line 21: curious how we decided on the value for what a good max duration would be

vinay-gopalan (approved):
Nice! LGTM, had one question on how the max duration for rollback is decided

path_rotate.go
benashz: I think those are based on the previous backoff implementation.

path_rotate.go
```diff
@@ -181,26 +187,32 @@ func (b *backend) pathRotateRoleCredentialsUpdate(ctx context.Context, req *logi
 	return nil, nil
 }
 
-// rollBackPassword uses naive exponential backoff to retry updating to an old password,
+// rollBackPassword uses exponential backoff to retry updating to an old password,
```
benashz, line 190: nit: suggest renaming to `rollbackPassword`

path_rotate_test.go
```diff
@@ -198,3 +204,88 @@ func TestManualRotate(t *testing.T) {
...
+
+	ctx := context.Background()
+
+	// works if the client always succeeds
+	fclient.count = 0
+	fclient.maxCount = 0
+	fclient.password = "password"
+	err := b.rollBackPassword(ctx, cfg, "old")
+	assert.Nil(t, err)
+	assert.Equal(t, "old", fclient.password)
```
benashz, line 264: Can we also assert something about `fclient.count`?

path_rotate_test.go
```diff
@@ -198,3 +204,88 @@ func TestManualRotate(t *testing.T) {
...
+		LDAP: &client.Config{
+			ConfigEntry: &ldaputil.ConfigEntry{
+				BindDN: "",
+			},
+		},
+	}
+
+	ctx := context.Background()
+
+	// works if the client always succeeds
```
benashz, line 258: I love the tests here! I think converting this to be a table driven test might be a cleaner approach. We could get a fake client for each sub test.

path_rotate_test.go
```diff
@@ -198,3 +204,88 @@ func TestManualRotate(t *testing.T) {
...
+	panic("nope")
+}
+
+func (f *failingRollbackClient) Execute(conf *client.Config, entries []*ldif.Entry, continueOnError bool) error {
+	panic("nope")
+}
+
+var _ ldapClient = (*failingRollbackClient)(nil)
+
+func TestRollback(t *testing.T) {
```
benashz, line 233: nit: consider renaming to `Test_rollbackPassword` or `TestRollbackPassword` to match the name of the function being tested here.

benashz (approved):
Nice catch! I added a few nits/suggestions for your consideration. After that :+1:

path_rotate.go
swenson: Yep, I didn't want to go significantly higher than the previous max, which was 81 seconds.

swenson:
Thanks!

PR #102, opened 2024-05-03
add hierarchical path handling to static role endpoints

fairclothjm:
This PR adds hierarchical path handling to the following APIs:
- [Static roles](https://developer.hashicorp.com/vault/api-docs/secret/ldap#static-roles)
- [Static role passwords](https://developer.hashicorp.com/vault/api-docs/secret/ldap#static-role-passwords)
- [Manually rotate static role password](https://developer.hashicorp.com/vault/api-docs/secret/ldap#manually-rotate-static-role-password)

This allows creating a static role name with an arbitrary number of forward slashes. For example,

```
$ vault write ldap/static-role/org/platform/dev \
    username="user3" \
    rotation_period="5m"
```

Where `org/platform/dev` is the role name. Creds can be read and rotated using the same role name and the respective API's. For example,

```
$ vault read ldap/static-cred/org/secure
Key                    Value
---                    -----
dn                     n/a
last_password          a3sQ6OkmXKt2dtx22kAt36YLkkxLsg4RmhMZCLYCBCbvvv67ILROaOokdCaGPEAE
last_vault_rotation    2024-05-03T16:39:27.174164-05:00
password               ECf7ZoxfDxGuJEYZrzgzTffSIDI4tx5TojBR9wuEGp8bqUXbl4Kr9eAgPjmizcvg
rotation_period        5m
ttl                    4m58s
username               user2

$ vault write -f ldap/rotate-role/org/secure
```

Most importantly, this allows us to perform LIST operations to query the available roles. For example,

```
$ vault list ldap/static-role/org/
Keys
----
platform/
secure

$ vault list ldap/static-role/org/platform
Keys
----
dev
```

### DIFF
--- a/bootstrap/configure.sh
+++ b/bootstrap/configure.sh
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
--- a/bootstrap/configure.sh
+++ b/bootstrap/configure.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+PLUGIN_DIR=$1
+PLUGIN_NAME=$2
+PLUGIN_PATH=$3
+
+echo "==> PLUGIN_DIR: $PLUGIN_DIR"
+echo "==> PLUGIN_NAME: $PLUGIN_NAME"
+echo "==> PLUGIN_PATH: $PLUGIN_PATH"
+
+# Try to clean-up previous runs
+vault secrets disable "${PLUGIN_PATH}"
+vault plugin deregister secret "${PLUGIN_NAME}"
+killall "${PLUGIN_NAME}"
+
+# Copy the binary so text file is not busy when rebuilding & the plugin is registered
+cp ./bin/"$PLUGIN_NAME" "$PLUGIN_DIR"
+
+# Sets up the binary with local changes
+vault plugin register \
+    -sha256="$(shasum -a 256 "$PLUGIN_DIR"/"$PLUGIN_NAME" | awk '{print $1}')" \
+    -version="0.0.1" \
+    secret "${PLUGIN_NAME}"
+
+if [ -e scripts/custom.sh ]
+then
+  . scripts/custom.sh
+fi
+
--- a/path_rotate_test.go
+++ b/path_rotate_test.go
@@ -95,7 +95,7 @@ func TestManualRotateRole(t *testing.T) {
 		configureOpenLDAPMount(t, b, storage)
 		createRole(t, b, storage, roleName)
 
-		resp, _ := readStaticCred(t, b, storage, roleName)
+		resp := readStaticCred(t, b, storage, roleName)
 
 		if resp.Data["password"] == "" {
 			t.Fatal("expected password to be set, it wasn't")
@@ -114,7 +114,7 @@ func TestManualRotateRole(t *testing.T) {
 			t.Fatalf("err:%s resp:%#v\n", err, resp)
 		}
 
-		resp, _ = readStaticCred(t, b, storage, roleName)
+		resp = readStaticCred(t, b, storage, roleName)
 
 		if resp.Data["password"] == "" {
 			t.Fatal("expected password to be set after rotate, it wasn't")
@@ -142,7 +142,7 @@ func TestManualRotateRole(t *testing.T) {
 		passwords := make([]string, 0)
 		// rotate all the creds
 		for _, role := range roles {
-			resp, _ := readStaticCred(t, b, storage, role)
+			resp := readStaticCred(t, b, storage, role)
 
 			if resp.Data["password"] == "" {
 				t.Fatal("expected password to be set, it wasn't")
@@ -161,7 +161,7 @@ func TestManualRotateRole(t *testing.T) {
 				t.Fatalf("err:%s resp:%#v\n", err, resp)
 			}
 
-			resp, _ = readStaticCred(t, b, storage, role)
+			resp = readStaticCred(t, b, storage, role)
 
 			newPassword := resp.Data["password"]
 			if newPassword == "" {
--- a/path_rotate_test.go
+++ b/path_rotate_test.go
@@ -10,14 +10,15 @@ import (
 	"time"
 
 	"github.com/go-ldap/ldif"
+	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/vault-plugin-secrets-openldap/client"
 	"github.com/hashicorp/vault/sdk/helper/ldaputil"
 	"github.com/hashicorp/vault/sdk/logical"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestManualRotate(t *testing.T) {
-	t.Run("rotate root", func(t *testing.T) {
+func TestManualRotateRoot(t *testing.T) {
+	t.Run("happy path rotate root", func(t *testing.T) {
 		b, storage := getBackend(false)
 		defer b.Cleanup(context.Background())
 
@@ -83,23 +84,29 @@ func TestManualRotate(t *testing.T) {
 			t.Fatal("should have got error, didn't")
 		}
 	})
+}
 
-	t.Run("rotate role", func(t *testing.T) {
+func TestManualRotateRole(t *testing.T) {
+	t.Run("happy path rotate role", func(t *testing.T) {
 		b, storage := getBackend(false)
 		defer b.Cleanup(context.Background())
 
-		data := map[string]interface{}{
-			"binddn":      "tester",
-			"bindpass":    "pa$$w0rd",
-			"url":         "ldap://138.91.247.105",
-			"certificate": validCertificate,
+		roleName := "hashicorp"
+		configureOpenLDAPMount(t, b, storage)
+		createRole(t, b, storage, roleName)
+
+		resp, _ := readStaticCred(t, b, storage, roleName)
+
+		if resp.Data["password"] == "" {
+			t.Fatal("expected password to be set, it wasn't")
 		}
+		oldPassword := resp.Data["password"]
 
 		req := &logical.Request{
-			Operation: logical.CreateOperation,
-			Path:      configPath,
+			Operation: logical.UpdateOperation,
+			Path:      rotateRolePath + roleName,
 			Storage:   storage,
-			Data:      data,
+			Data:      nil,
 		}
 
 		resp, err := b.HandleRequest(context.Background(), req)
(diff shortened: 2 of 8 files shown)


path_static_roles.go
```diff
@@ -20,10 +21,24 @@ const (
 	staticRolePath = "static-role/"
 )
 
+// GenericNameWithForwardSlashRegex is a regex which requires a role name. The role name can
+// include any number of alphanumberic characters separated by forward slashes.
+func GenericNameWithForwardSlashRegex(name string) string {
```
vinay-gopalan, line 26: It would be great if we could have a small unit test validating this regex does what we expect. I personally am not great with regexes and usually am unable to tell if regexes run as expected 😅 

Disregard if this is not a concern! :)

path_static_roles.go
```diff
@@ -20,10 +21,24 @@ const (
...
+	return fmt.Sprintf("(/(?P<%s>\\w(([\\w-./]+)?\\w)?))", name)
+}
+
+// OptionalGenericNameWithForwardSlashListRegex is a regex for optionally including a
+// role path in list options. The role path can be used to list nested roles at
+// arbitrary depth.
+func OptionalGenericNameWithForwardSlashListRegex(name string) string {
+	// TODO(JM): List regex should support optional trailing slash
```
vinay-gopalan, line 34: I see this is still a TODO referred to in the tests. Will it be addressed in a follow up PR?

path_static_roles_test.go
```diff
@@ -707,6 +479,47 @@ func TestListRoles(t *testing.T) {
...
+			resp, err := createStaticRoleWithData(t, b, storage, role, data)
+			if err != nil || (resp != nil && resp.IsError()) {
+				t.Fatalf("err:%s resp:%#v\n", err, resp)
+			}
+		}
+
+		// TODO(JM): List regex should support optional trailing slash
+		// rolePaths := []string{"org", "org/", "org/platform", "org/platform/"}
+		rolePaths := []string{"org/", "org/platform/"}
+		for _, rolePath := range rolePaths {
```
vinay-gopalan, line 503: Does this test set the new `path` field? It looks like the `Data` being passed to it is `nil`, should we add a test step that also uses the new field?

vinay-gopalan (commented):
PR is looking good! Had a couple initial questions

bootstrap/configure.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
benashz, line 1: out of curiosity is this boiler plate that comes from a know source? If so, it would be great to provide a link to it, in the case it ever needs to be updated.

path_rotate_test.go
```diff
@@ -83,107 +84,99 @@ func TestManualRotate(t *testing.T) {
...
+		configureOpenLDAPMount(t, b, storage)
+		createRole(t, b, storage, roleName)
 
-		req := &logical.Request{
-			Operation: logical.CreateOperation,
-			Path:      configPath,
-			Storage:   storage,
-			Data:      data,
-		}
+		resp, _ := readStaticCred(t, b, storage, roleName)
```
benashz, line 98: should fail if `resp` is nil, I think if we don't we'll get  nil pointer exception further down

path_static_creds_test.go
```diff
@@ -14,105 +14,47 @@ func TestCreds(t *testing.T) {
...
 			"dn":              "uid=hashicorp,ou=users,dc=hashicorp,dc=com",
-			"rotation_period": "60s",
-		}
-
-		req = &logical.Request{
-			Operation: logical.CreateOperation,
-			Path:      staticRolePath + "hashicorp",
-			Storage:   storage,
-			Data:      data,
+			"rotation_period": float64(60),
```
benashz, line 22: do we nee to worry about no longer doing the duration parse here?

path_static_creds_test.go
```diff
@@ -121,3 +63,47 @@ func TestCreds(t *testing.T) {
...
+		Data:      nil,
+	}
+
+	return b.HandleRequest(context.Background(), req)
+}
+
+func assertReadStaticCred(t *testing.T, b *backend, storage logical.Storage, roleName string, data map[string]interface{}) {
+	t.Helper()
+	resp, err := readStaticCred(t, b, storage, roleName)
+	if err != nil || (resp != nil && resp.IsError()) {
```
benashz, line 81: do we need to fail if `resp` is `nil`?

path_static_creds_test.go
```diff
@@ -121,3 +63,47 @@ func TestCreds(t *testing.T) {
...
+		t.Fatalf("err:%s resp:%#v\n", err, resp)
+	}
+
+	if resp.Data["dn"] != data["dn"] {
```
benashz, line 85: Should we check to see if data has `dn`?

path_static_creds_test.go
```diff
@@ -121,3 +63,47 @@ func TestCreds(t *testing.T) {
...
+		t.Fatalf("expected dn to be %s but got %s", data["dn"], resp.Data["dn"])
+	}
+
+	if resp.Data["password"] == "" {
+		t.Fatal("expected password to be set, it wasn't")
+	}
+
+	if resp.Data["username"] != data["username"] {
```
benashz, line 93: similar question about the need to check if the data map contains the key.

path_static_creds_test.go
```diff
@@ -121,3 +63,47 @@ func TestCreds(t *testing.T) {
...
+		t.Fatalf("expected username to be %s but got %s", data["username"], resp.Data["username"])
+	}
+
+	if resp.Data["last_vault_rotation"] == nil {
+		t.Fatal("expected last_vault_rotation to be set, it wasn't")
+	}
+
+	expected := data["rotation_period"].(float64)
```
benashz, line 101: we may bet a panic on unhandled type assert errors

path_static_roles.go
```diff
@@ -43,7 +64,7 @@ func (b *backend) pathListStaticRoles() []*framework.Path {
 func (b *backend) pathStaticRoles() []*framework.Path {
 	return []*framework.Path{
 		{
-			Pattern: staticRolePath + framework.GenericNameRegex("name"),
+			Pattern: strings.TrimSuffix(staticRolePath, "/") + GenericNameWithForwardSlashRegex("name"),
```
benashz, line 67: nit, it is a bit odd that we trimming a constant but I guess it is easier than decomposing `staticRolePath` ?

path_static_roles_test.go
```diff
@@ -489,38 +364,14 @@ func TestRoles(t *testing.T) {
...
 
-		req = &logical.Request{
-			Operation: logical.CreateOperation,
-			Path:      staticRolePath + "hashicorp",
-			Storage:   storage,
-			Data:      data,
-		}
-
-		resp, err = b.HandleRequest(context.Background(), req)
+		resp, _ := createStaticRoleWithData(t, b, storage, "hashicorp", data)
```
benashz, line 374: any reason not to handle the error here?

path_static_roles_test.go
```diff
@@ -530,39 +381,15 @@ func TestRoles(t *testing.T) {
...
```
benashz, line 392: any reason not to handle the error here?

path_static_roles_test.go
```diff
@@ -767,16 +580,12 @@ func TestWALsDeletedOnRoleCreationFailed(t *testing.T) {
...
-				"dn":              "uid=hashicorp,ou=users,dc=hashicorp,dc=com",
-				"rotation_period": "5s",
-			},
-		})
+		data := map[string]interface{}{
+			"username":        "hashicorp",
+			"dn":              "uid=hashicorp,ou=users,dc=hashicorp,dc=com",
+			"rotation_period": "5s",
+		}
+		resp, err := createStaticRoleWithData(t, b, storage, "hashicorp", data)
```
benashz, line 588: Should `resp` be nil in this case? If so we may want to assert that below.

benashz (commented):
Looking good. I had a few minor suggestions, questions, and nits that you might want to address.

bootstrap/configure.sh
fairclothjm: It's just a bash shebang that my editor adds automatically when I create a new shell script. I am happy to remove it if it causes confusion. I believe it should work on any modern system. https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang

path_rotate_test.go
fairclothjm: Since that code path is being tested in path_static_creds_test.go, I opted to not test it here. You are right about the panic though so I can make the change.

path_static_roles_test.go
fairclothjm: The `path` field is extracted from the request path by the framework. If it wasn't set properly, this test would fail

path_static_roles.go
fairclothjm: I planned to add tests for this but it slipped my mind. I will add some. Eventually, I plan to move this to the Vault sdk but for now we will embed them in the plugin

path_static_roles.go
```diff
@@ -20,10 +21,24 @@ const (
...
+// GenericNameWithForwardSlashRegex is a regex which requires a role name. The role name can
+// include any number of alphanumberic characters separated by forward slashes.
+func GenericNameWithForwardSlashRegex(name string) string {
+	return fmt.Sprintf("(/(?P<%s>\\w(([\\w-./]+)?\\w)?))", name)
+}
+
+// OptionalGenericNameWithForwardSlashListRegex is a regex for optionally including a
+// role path in list options. The role path can be used to list nested roles at
+// arbitrary depth.
+func OptionalGenericNameWithForwardSlashListRegex(name string) string {
```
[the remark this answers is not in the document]
fairclothjm: will do!

path_static_roles.go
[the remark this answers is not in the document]
fairclothjm: Yes, I think so. I am deferring this until I get all of the API's updated. I wasn't able to figure out how to make the trailing slash optional but I think it is required as per the docs on the `Pattern` field in `framework.Path`

fairclothjm: Yes, I think so. I am deferring this until I get all of the API's updated. I wasn't able to figure out how to make the trailing slash optional but I think it is required as per the docs on the `Pattern` field in `framework.Path`

path_static_roles_test.go
fairclothjm: It was being tested elsewhere, but I can handle it in the helper

fairclothjm:
@benashz @vinay-gopalan I have addressed the bulk of your comments. Any remaining test improvements I will do once we get past feature freeze.

path_static_roles.go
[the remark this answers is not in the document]
fairclothjm: I will add this in a follow up PR

vinay-gopalan (approved):
LGTM! Thanks for all the clarifying comments, I am fine with additional test improvements being handled after feature freeze

PR #123, opened 2024-10-21
prevent role and set duplicates

fairclothjm:
There is an edge case where a user can add an LDAP user or service account to more than one role or set. This is only possible after the backend has been reloaded.

Roles and sets can be defined as hierarchical paths (i.e. a graph). So we update the initialization path to perform a breadth-first search on the plugin's storage for both roles and sets. If we find an entry, we add it to the managed users map.

We attempt to make as few reads from storage as possible by returning the static roles from the `loadManagedUsers` func and passing them to the `initQueue` function instead of performing another lookup of each node in the graph. We do not want superfluous storage access on plugin initialization which could degrade Vault startup times.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -5,6 +5,7 @@ package openldap
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


rotation.go
```diff
@@ -585,50 +572,3 @@ func (b *backend) popFromRotationQueueByKey(name string) (*queue.Item, error) {
 	}
 	return nil, queue.ErrEmpty
 }
-
-// getAllStaticRoles returns a slice of all static roles from storage
-func (b *backend) getAllStaticRoles(ctx context.Context, s logical.Storage) ([]string, error) {
```
fairclothjm, line 590: This was refactored to the generic `walkStoragePath` to be used for either roles or sets

backend.go
```diff
@@ -149,10 +151,57 @@ type backend struct {
...
+		entryExists, err := walker(key)
+		if err != nil {
+			return fmt.Errorf("unable to read entry %q: %w", key, err)
+		}
+
+		if !entryExists && strings.HasSuffix(key, "/") {
+			// this is a directory
+			subKeys, err := s.List(ctx, path+key)
+			if err != nil {
+				return fmt.Errorf("unable to list keys: %w", err)
```
tomcf-hcp, line 183: Is there ever a case where "foo/bar" would fail, but the "foo/bar/baz" would succeed and both of those would be "correct"?

fairclothjm: An error returned from `storage.List()` means we could not access storage or we are in some bad state with the keys and we are dead in the water.

We have tests to cover the cases that I think you are describing:
https://github.com/hashicorp/vault-plugin-secrets-openldap/blob/156b280cd38c4d77e61d34839dad42396f9384f1/rotation_test.go#L48-L55

vinay-gopalan (approved):
LGTM!

PR #156, opened 2025-03-31
Fix issue where roles created before 0.14.5 had a nil NextVaultRotation value #VAULT-35117

elliesterner:
There was an issue with static roles that opted to enable `skip_import_rotation` that was fixed in [this PR](https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/140). However, if the role was created after the fix was introduced, it's `NextVaultRotation` was never set, so vault rotates the credential on startup (queue initialization). 

This includes a fix for 2 other issues:
- `NextVaultRotation` was not getting set on update, so if rotation_period was change, the TTL would not be updated
- On update, vault would prematurely rotate the credential of a static role with `skip_import_rotation=true` before it's TTL. 

# Overview
A high level description of the contribution, including:
Who the change affects or is for (stakeholders)?
What is the change?
Why is the change needed?
How does this change affect the user experience (if at all)?

- [ ] [Issue #1234](https://github.com/hashicorp/vault/issues/1234)
- [ ] [PR #1234](https://github.com/hashicorp/vault/pr/1234)

# Contributor Checklist
- [ ] Backwards compatible


rotation_test.go
```diff
@@ -248,6 +248,53 @@ func TestAutoRotate(t *testing.T) {
...
+		}
+		if err := storage.Put(context.Background(), entry); err != nil {
+			t.Fatal("failed to write role to storage", err)
+		}
+
+		// Reload backend to similate a Vault restart/startup memory state
+		getBackendWithConfig(config, false)
+
+		// TODO: this is hacky because the queue ticker runs every 5 seconds
+		time.Sleep(8 * time.Second)
```
elliesterner, line 289: Is there a better way to make sure the periodic ticker that triggers credential rotation runs? I wonder if the other tests that rely on `getBackendWithConfig` are not working as intended??

path_static_roles.go
```diff
@@ -458,6 +458,22 @@ type staticAccount struct {
...
+	// created withskip_import_rotation set, we need to use the current time
+	// instead of LastVaultRotation because LastVaultRotation is 0
+	// This situation was fixed by https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/140.
+	if s.NextVaultRotation.IsZero() {
+		// Previously skipped import rotation roles had a LastVaultRotation value of zero
+		if s.LastVaultRotation.IsZero() {
+			s.SetNextVaultRotation(time.Now())
+		}
+
+		s.SetNextVaultRotation(s.LastVaultRotation)
```
davidadeleon, line 474: Perhaps I am misunderstanding the intent here, but this seems as though if `s.NextVaultRotation` is zero  AND `s.LastVaultRotation` is zero we'll end up setting `s.NextVaultRotation` to zero regardless here.

path_static_roles.go
```diff
@@ -458,6 +458,23 @@ type staticAccount struct {
 // NextRotationTime calculates the next rotation by adding the Rotation Period
 // to the last known vault rotation
 func (s *staticAccount) NextRotationTime() time.Time {
+	// If an account's NextVaultRotation period is nil, it means that the
```
fairclothjm, line 461:
```suggestion
	// If an account's NextVaultRotation period is zero, it means that the
```

path_static_roles.go
```diff
@@ -458,6 +458,23 @@ type staticAccount struct {
...
+	// role was created before we added the `NextVaultRotation` field. In this
+	// case, we need to calculate the next rotation time based on the
+	// LastVaultRotation and the RotationPeriod. However, if the role was
+	// created withskip_import_rotation set, we need to use the current time
```
fairclothjm, line 465:
```suggestion
	// created with skip_import_rotation set, we need to use the current time
```

rotation.go
```diff
@@ -55,6 +55,30 @@ func (b *backend) populateQueue(ctx context.Context, s logical.Storage, roles ma
 			continue
 		}
 
+		// If an account's NextVaultRotation period is nil, it means that the
+		// role was created before we added the `NextVaultRotation` field. In this
+		// case, we need to calculate the next rotation time based on the
+		// LastVaultRotation and the RotationPeriod. However, if the role was
+		// created withskip_import_rotation set, we need to use the current time
```
vinay-gopalan, line 62:
```suggestion
		// created with skip_import_rotation set, we need to use the current time
```

rotation_test.go
```diff
@@ -248,6 +251,104 @@ func TestAutoRotate(t *testing.T) {
...
+
+		// check if first rotation is different from second rotation
+		if !firstRotation.Equal(secondRotation) {
+			t.Fatal("expected first rotation to be equal to second rotation to prove that credential wasnt rotated")
+		}
+	})
+
+	// This is to test static roles created before 0.14.5
+	// In 0.14.5, vault started persisting `NextVaultRotation`
+	t.Run("zero NextVaultRotation for a static role with `skip_import_rotation` does not cause rotate after backend reload", func(t *testing.T) {
```
vinay-gopalan, line 306: What is the difference between these two tests? Seems like they both set `skip_import_rotation` on a static role, zero out `NextVaultRotation`, reload the backend, and ensure that the credential was not rotated?

vinay-gopalan (approved):
Looks great! Had a couple non-blocking comments, but LGTM otherwise!

rotation_test.go
elliesterner: Just the flag `skip_import_rotation`. It's testing this scenario: https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/156/files#diff-22918e175431702db9dac2c03b2350a68ed7ab571fe03a3b326c73572a3718e4R69

vinay-gopalan (approved):

rotation_test.go
```diff
@@ -248,6 +251,188 @@ func TestAutoRotate(t *testing.T) {
...
+
+		_, err := createStaticRoleWithData(t, b, storage, roleName, data)
+		require.NoError(t, err)
+
+		originalRole, err := b.staticRole(context.Background(), storage, roleName)
+		if err != nil {
+			t.Fatal("failed to fetch static role", err)
+		}
+
+		// Update static role's rotation period to 1h
```
vinay-gopalan, line 319: nit: 
```suggestion
		// Update static role's rotation period to 5m
```

rotation_test.go
vinay-gopalan: Gotcha, thanks!

vinay-gopalan (dismissed):

vinay-gopalan (approved):

PR #170, opened 2025-05-19
Add RACF Passphrase Support

jaireddjawed:
# Overview

This PR adds RACF Passphrase support as an additional auth approach to the existing RACF password.

# Contributor Checklist
- [ ] Backwards compatible


client/fieldregistry.go
```diff
@@ -62,6 +62,7 @@ type fieldRegistry struct {
 	PasswordLastSet    *Field `ldap:"passwordLastSet"`
 	RACFID             *Field `ldap:"racfid"`
 	RACFPassword       *Field `ldap:"racfPassword"`
+	RACFPassphrase     *Field `ldap:"racfPassphrase"`
```
tomcf-hcp, line 65: Payman pointed out that the [RACF documents](https://www.ibm.com/docs/en/zos/2.4.0?topic=sob-changing-user-password-password-phrase-in-racf-using-sdbm) uses `racfPassPhrase` (camel case). We should match that convention where it makes sense.

client/schema.go
```diff
@@ -37,10 +37,17 @@ func GetSchemaFieldRegistry(schema string, newPassword string) (map[*Field][]str
 		fields := map[*Field][]string{FieldRegistry.UserPassword: {newPassword}}
 		return fields, nil
 	case SchemaRACF:
-		fields := map[*Field][]string{
-			FieldRegistry.RACFPassword:   {newPassword},
-			FieldRegistry.RACFAttributes: {"noexpired"},
+		fields := map[*Field][]string{}
+
+		// passphrase is used instead when the password string is longer than 8 characters
+		if len(newPassword) > 8 {
```
tomcf-hcp, line 43: I'm worried about this. What happens to the state of RACF if someone starts with a short password, and then makes a longer passphrase (or vice versa)? Will both exist on the RACF side?
This is hiding state that makes a difference in RACF, and will lead to more state management to worry about and go wrong than your original implementation I think?

vinay-gopalan:
FYI going to punt this to the next plugin release. This will not be making it into the Vault 1.20 cutoff for the plugin that will be going out today

client/schema.go
jaireddjawed: I see the issue now. Yes, if someone were to enter a short password and then a long one, the old field will still be set. I'll set the opposite field to be

PR #189, opened 2025-08-12
Vault 38807 plugin quality configure plugin and test library api crud with enos

HamzaShili65:
Adds a Terraform module and testing script to validate the API for the library endpoint. It integrates with Enos by adding a test CRUD API step that runs after successful plugin configuration in `enos-scenario-openldap-smoke.hcl` and `enos-scenario-openldap-restart.hcl`.


fairclothjm (approved):
LGTM!

PR #191, opened 2025-08-13
Vault 38809 plugin quality test configure plugin and vault leader change with enos

HamzaShili65:
Add `enos-scenario-openldap-leader_change.hcl` to test that the Vault OpenLDAP secrets engine plugin works correctly after a leader change.

This scenario creates a Vault cluster with the OpenLDAP secrets engine plugin installed and configured, and starts an OpenLDAP server. It then tests the plugin by creating static and dynamic roles, verifying that they can be created, read, updated, and deleted via the Vault API. After that, it forces a Vault leader stepdown followed by a leader election and verifies that the plugin still works correctly after the leader change

command to run the scenario:
`enos scenario launch openldap_leader_change arch:amd64 artifact_source:artifactory artifact_type:package config_mode:env distro:amzn edition:ce ip_version:4 seal:shamir ldap_artifact_source:local `


fairclothjm (approved):
LGTM!

PR #193, opened 2025-08-15
Vault 38808 plugin quality test configure plugin and test config endpoint root rotation

HamzaShili65:
Add tests for manual root rotation and rotation manager root rotations.


fairclothjm (approved):
LGTM

PR #212, opened 2025-09-22
Enable self credentials rotation for static roles

jadeidev:
# Overview
### Who the change affects or is for (stakeholders)?
This change is for Vault users and administrators who manage static LDAP accounts and require Vault to rotate credentials using the account’s own password, rather than a privileged bind DN. Stakeholders include security teams and operators who need tighter control over credential rotation for sensitive accounts.

### What is the change?
This contribution enables support for "self-managed" static accounts in the Vault OpenLDAP secrets engine. When enabled, Vault rotates credentials by authenticating as the managed account itself, using its current password, rather than relying on a privileged bind DN.
This feature introduces new attributes to the static role schema: 

- `self_managed` (default: `false`) parameter is optional and indicates whether the role manages its own password rotation. If true, Vault will perform rotations by authenticating as this account using its current password (no privileged bind DN).
This requires the "password" parameter to be set on creation, and the "dn" parameter to be set as well. 
This field is immutable after creation. If false (the default), Vault will use the configured bind DN to perform rotations.

- `password` parameter is required only if "self_managed" is true, and configures the current password for the entry. 
This allows Vault to assume management of an existing account. The password will be rotated on creation unless 
the `skip_import_rotation` parameter is set to true. The password is not returned in read operations.

- `self_managed_max_invalid_attempts` parameter is optional and configures the maximum number of invalid current-password attempts for self-managed accounts. A value equal to 0 means use the default (5), and a negative value means unlimited attempts. When the maximum number of attempts is reached, automatic rotation is suspended until the role is updated via the `password` parameter. This field is immutable after creation.

- `rotation_suspended` field is only returned in read operations and indicates whether automatic rotation is currently suspended for this static account due to too many invalid current-password attempts on self-managed accounts.

### Why is the change needed?
Some environments restrict privileged access or require that password changes be performed by the account owner. Supporting self-managed rotation allows Vault to manage credentials for accounts where privileged bind is not possible or not desired, improving flexibility and security.

### How does this change affect the user experience (if at all)? 
Users can now create static roles with the `self_managed` option, providing the current password and DN. Vault will rotate credentials using the account’s own authentication, and will suspend automatic rotation after a configurable number of failed attempts, requiring manual intervention. This enhances Vault’s ability to manage a wider range of LDAP accounts and improves compliance with organizational policies.

# Contributor Checklist
- [ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet - **not sure where to add it. please let me know ill be happy to do it**
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests) - all tests run in CI
- [x] Backwards compatible
- [x] Changelog entry added. See [Updating the Changelog](https://github.com/hashicorp/vault-plugin-secrets-openldap/blob/main/README.md#updating-the-changelog).

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.


jadeidev:
most of the code in this PR has been incoporated in the new Enterprise version of LDAP Secrets Engine (see vault 2.0 realse notes FEATURES https://github.com/hashicorp/vault/blob/main/CHANGELOG.md)

